Issue #56, opened 2020-05-13
Prevent dragging from moving the screen around on mobile


isaacHagoel:

isaacHagoel:
merged a fix just now

PR #57, opened 2020-05-14
56 - this is an attempt to fix the issue. it works on firefox emulato…

isaacHagoel:
…r. will test on a mobile device before merging


isaacHagoel:
resolves #56

Issue #64, opened 2020-05-26
Dispatch finalize event on the origin dndzone if element was dropped elsewhere


isaacHagoel:
Currently only the destination zone gets a finalize event when the element in dropped. The origin dndzone should get one too (if it is not also the destination) so that it 'knows' that the element was permanently removed.

PR #65, opened 2020-05-27
64 - dispatching finalize event on origin dz when element is dropped …

isaacHagoel:
…into a different dz


isaacHagoel:
fixes #64

Issue #81, opened 2020-06-15
Customize the dropzone style?


nathancahill:
Right now the style is hardcoded as `dz.style.outline = 'rgba(255, 255, 102, 0.7) solid 2px'`

Would you be interested in a PR that customizes that style?

isaacHagoel:
Sure. My thinking was to keep things as simple as possible and add customisation options as needs arise 
Do you have something specific in mind (API)?

nathancahill:
This is the idea: https://github.com/isaacHagoel/svelte-dnd-action/compare/master...nathancahill:master

nathancahill:
Basically: 

```js
use:dndzone={{items, activeStyle: { outline: 'rgba(255, 255, 102, 0.7) solid 2px' } }}
```

isaacHagoel:
The code looks legit! nice work sir.
I do have 4 comments:

1. JSDoc will need to have the extra param added (easy)
2. README.md should include the new option
3. I wonder whether there is a clearer name (from the user perspective) than 'activeStyle' for the option... not sure there is tbh, raising it anyway
4. Removing the active styles might need to be more sophisticated. I thought I could get away with setting the outline property to '' because it is rarely used. What if we set background or border to ''? That might result in unexpected behaviour if there was another value there before. Do you think we should backup the original values and restore them? Alternatively we should be explicit about this in the README

nathancahill:
3. Yes, the name can be better. Maybe `considerStyle`? `active` is vague.
 
4. The styles are only being removed if applied directly to the element with the inline styles tag. It's not wiping styles applied Svelte's `<style>` or anything else. Still better to backup and restore.

Adding a fifth item as I've been playing with the code more:

5. The `dzToConfig` logic means that dropzones are getting styled with the config from the dropzone the item is being dragged *from*, this might be different. I'll try to fix that today.

isaacHagoel:
3. it doesn't trigger on consider but when an element from the same type is being dragged around. Maybe something like "dropTargetStyle" or "activeDropTargetStyle"?
4. You are right. I am aware. If backing up is too tricky it is okay to not do it as long as we are explicit about it in the README (we can improve it later + it is a user defined option)
5. Good point! Each dz should be styled based on its own config.  

Adding 6:
Should the existence of this option cancel the default highlighting or add to it?

PlkMarudny:
[...]

PR #82, opened 2020-06-16
Add dropTargetStyle

nathancahill:
Per #81:

- [x] JSDoc added
- [x] README.md updated
- [x] Uses `dropTargetStyle` for the name
- [ ] Backup styles and reapply them
- [x] Each dnd-zone is styled by it's own config
- [x] Existence of the option overrides the default styles


nathancahill:
I think currently the backing up styles and reapplying them gets complex (creating a style cache per dnd-zone, refreshing the cache is styles are changed by Svelte outside of dnd, reapplying the cached styles).

So I'll add a note to the README.

isaacHagoel:
Overall, looks good. Nice solution with passing in the function :)
JSDOC should probably explain what kind of function is expected but I can add that in later.
I'll test it out, hopefully later today, and if everything works as expected (i think it will be), I will merge. Thank you sir!

Issue #106, opened 2020-08-18
Draggable is jumping when first entering the dropzone. Question on shadow elements.


MaddoxDevelopment:
Great library, thanks for making it.

For some reason, when I drag my element from one zone to another zone, it jumps back to the left side, do you know what might be causing this issue? 

The height of both drop-zones are 100vh.

Example Video: https://streamable.com/bm0sue

Also, is there a way to change the HTML of the element being dragged? I would like to transform it to look like the input that is going to be rendered, but cannot figure out a good way to intercept that. 

Thanks!

isaacHagoel:
Hi,
This is actually the intended behaviour. If you add a border around the element it might make more visual sense.
The dragged element morphs to match its new shape in the container underneath (which is much larger in your case). It morphs around the mouse pointer such that the pointer maintains its relative position within the element (ex: if it was next to the top left corner before morphing it would stay there after morphing). 
Not doing so, can cause the pointer to end up outside of the new shape.

The code responsible for the morphing (and maintaining the pointer position) is [here](https://github.com/isaacHagoel/svelte-dnd-action/blob/2b0983743a3ba26e20708d258dfcae33192af168/src/helpers/styler.js#L52).
You can see which computed styles are copied and which ones aren't. 

In order to make it look like the input additional styles would need to be copied (and it might need to be an input element to begin with).

MaddoxDevelopment:
Oh i see, I added a border to the dragged element and it makes much more sense visually, i'll figure out a way to style it then. Thanks for your help!

MaddoxDevelopment:
Re-opening this for a further request.

I pulled the repo to try out a few things and figured out I can solve all of my issues by having access to the inner html that is being set on the dragged element in the morph function. Would this be something you'd be willing to expose via the api? Possibly some sort of configuration to allow us to pass markup that the morphed element gets turned into when it gets picked up.

<img width="1226" alt="Screen Shot 2020-08-18 at 7 34 52 PM" src="https://user-images.githubusercontent.com/11528924/90578842-e8c1b700-e189-11ea-955a-c4c7673ea7ae.png">

Just changed the bottom line there and it works great. https://streamable.com/7xonif

@isaacHagoel

isaacHagoel:
Thanks for the PR. I will have a look at it and test hopefully today (if not then tomorrow for sure). 
Yeah, I'm totally open to this.

PR #107, opened 2020-08-19
Add ability to intercept the dragged element.

MaddoxDevelopment:
This adds a hook to the config to be able to intercept the dragged element while it is being morphed. This allows you to do things such as modifying the HTML of the dragged element.

Example Code:
```
<div
        use:dndzone={{ items: blocks, flipDurationMs: 300,
        transformDraggedElement : (el) => {
          el.innerHTML = "<label>New Field</label><input class='form-control shadow'/>";
          return el
        },
        dropFromOthersDisabled: true, dropTargetStyle: { outline: 'white solid 0px' } }}
        on:consider={handler}
        on:finalize={handler}
>
```

Result: https://streamable.com/zca69w


MaddoxDevelopment:
Edit: I think I have this on the wrong method, needs to go on createDraggedElementFrom. Updating now.

src/action.js
```diff
@@ -292,7 +292,7 @@ export function dndzone(node, options) {
         shadowElData = {...draggedElData, isDndShadowItem: true};
 
         // creating the draggable element
-        draggedEl = createDraggedElementFrom(originalDragTarget);
+        draggedEl = createDraggedElementFrom(originalDragTarget, config.transformDraggedElement, currentIdx);
```
isaacHagoel, line 295: 1. This is the config of the dnd zone it is dragged from. shouldn't it be manipulated based on the config of the dnd zone it enters?
2. What are you using the origin index for in the transform function?

src/helpers/styler.js
```diff
@@ -31,6 +34,11 @@ export function createDraggedElementFrom(originalElement) {
     window.setTimeout(() => draggedEl.style.transition +=`, ${trs('top')}, ${trs('left')}`,0);
     draggedEl.style.zIndex = '9999';
     draggedEl.style.cursor = 'grabbing';
+
+    if(transformDraggedElement) {
```
isaacHagoel, line 38: why here and not where you originally had it (when morphing)? i might be misunderstanding the feature/ usecase

src/action.js
MaddoxDevelopment: I originally had it in the morph function, but it was being called for every single element that the draggable was going over and not being called for the last element in the list of items.

I am using the index to tell which item I dragged so I am able to change what the draggable looked like.

For instance:

```
[
      { id: randomString(), name: "string" },
      { id: randomString(), name: "combobox" },
      { id: randomString(), name: "spacer" },
    ]
```

If my items looks like this, I can check if index === 1 and then change the dragged element html to look like a combobox, etc.

MaddoxDevelopment: Here is how it functions now. I am able to use the index to change the draggable accordingly.

https://streamable.com/zca69w

src/helpers/styler.js
MaddoxDevelopment: ^ Check above comment

src/action.js
isaacHagoel: I am glad it works in this case but we need something that works generically for basically every situation the library can find itself in. Not sure why u had issues with the last item and such. I can probably help but...
Please help me understand the desired behaviour first. Feel free to correct any mistake I am making below. My understanding is:
- We want to define a function **per dndzone** and pass it in as config (as in dndzone1 might have f1 and dndzone2 might have f2)
- We expect the function to be called whenever - here I am not super clear. Here are some options, more than one may apply:
a. when the dragged element enters the dndzone (f1 if it enters dndzone1, f2 for dndzone2)
b. when the dragged element hovers over a new index after it had entered
c. when the dragged element leaves a dndzone (is there a reverse function?)
d. when drag start and the element is created? if so, it might not be consistent with the current intention of morphing - to make the dragged element look like it "belongs" in its would be destination
e. ?

MaddoxDevelopment: d. is the use case I was going for with it. This way you can have full control of how it looks when it is being dragged if needed. The current way of just cloning the node and changing styles is perfect if you just want to move the same "item" from one place to another, but it is lacking in terms of being able to customize that to display something completely different.

My original issue where the text was being resized because of the morph is a perfect example of this. The sidebar is small and you wouldn't want to display a textbox on it, but a preview of how it would look before dropped on the page is still needed.

This change doesn't restrict or cause any additional concerns that you would need to consider, as all it is doing is opening it up to the developer to apply additional styles to the draggable element as they please.

isaacHagoel: Transforming it when drag starts assumes there is only one other way it would need to look like, doesn't it? as in - it is assuming its destination and it is assuming the user doesn't intend to drop it back in its original zone

MaddoxDevelopment: Correct, even if the user does drop it back in its original dropzone, it doesn't keep those styles because it'll re-render with the updated item list.

isaacHagoel: I think this is too specific of a usecase. 
Let's take a step back for a minute. 
is what you really want is what you did initially but u changed to the current approach because you encountered some issues?

MaddoxDevelopment: I see now what you mean. It would make more sense for the zone that is being dropped onto to decide how it should change. It does make more sense in the morph function. I confused myself before by putting the function on the wrong dropzone.

I moved it to the morph function and it works well for transforming. The only thing I need is the index of the dragged item so I know what to transform it to.

isaacHagoel: maybe it can pass in the data of the item `draggedElData`?

MaddoxDevelopment: Yeah something like that would work, i'll take a look at it later. Do you plan on exposing any additional events such as onDragStart, onDragStop? From what I recall, consider and finalize are only invoked once you are over the dropzone.

isaacHagoel: consider is invoked on every change except for the final drop that invokes finalize. it doesn't include any information on why it was invoked but we could add it if needed (while trying to keep things as simple as possible)

README.md
```diff
@@ -85,14 +85,15 @@ npm install --save-dev svelte-dnd-action
...
+| `dropTargetStyle`         | Object<String> | No                                                           | `{outline: 'rgba(255, 255, 102, 0.7) solid 2px'}` | An object of styles to apply to the dnd-zone when items can be dragged in to it. Note: the styles override any inline styles applied to the dnd-zone. When the styles are removed, any original inline styles will be lost |
+| `transformDraggedElement` | Function       | No                                                           | `undefined`                                       | A function that is invoked when the draggable element is enters the dnd-zone. It passes in the dragged element as the first parameter, and the data of item from the items array as the second parameter. This allows you to override any fields on the element, such as changing the innerHTML. |
```
isaacHagoel, line 96: 1. `is enters` is probably a typo.
2. it is invoked every time the dragged element enters the list or hovers over a new index in the target list

src/action.js
```diff
@@ -331,6 +331,7 @@ export function dndzone(node, options) {
         dragDisabled = false,
         dropFromOthersDisabled = false,
         dropTargetStyle = DEFAULT_DROP_TARGET_STYLE,
+        transformDraggedElement,
```
isaacHagoel, line 334: please also add to the initialisation of config that starts at line 229. 
also please default to an empty function `()=>{}` to prevent issues if we try to invoke it elsewhere and the user didn't provide it

src/action.js
```diff
@@ -362,7 +365,7 @@ export function dndzone(node, options) {
             const draggableEl = node.children[idx];
             styleDraggable(draggableEl, dragDisabled);
             if (config.items[idx].hasOwnProperty('isDndShadowItem')) {
-                morphDraggedElementToBeLike(draggedEl, draggableEl, currentMousePosition.x, currentMousePosition.y);
+                morphDraggedElementToBeLike(draggedEl, draggableEl, currentMousePosition.x, currentMousePosition.y, config.transformDraggedElement, draggedElData);
```
isaacHagoel, line 368: draggedElData is not really needed inside (it is only passed through).
can we instead do something like:
```javascript
morphDraggedElementToBeLike(draggedEl, draggableEl, currentMousePosition.x, currentMousePosition.y, () => config.transformDraggedElement(draggedEl, draggedElData));
```
?

src/helpers/styler.js
```diff
@@ -69,10 +72,15 @@ export function morphDraggedElementToBeLike(draggedEl, copyFromEl, currentMouseX
     const computedStyle = window.getComputedStyle(copyFromEl);
     Array.from(computedStyle)
         .filter(s => s.startsWith('background') || s.startsWith('padding') || s.startsWith('font') || s.startsWith('text') || s.startsWith('align') ||
-        s.startsWith('justify') || s.startsWith('display') || s.startsWith('flex') || s.startsWith('border') || s === 'opacity' || s === 'color')
+            s.startsWith('justify') || s.startsWith('display') || s.startsWith('flex') || s.startsWith('border') || s === 'opacity' || s === 'color')
         .forEach(s =>
             draggedEl.style.setProperty(s, computedStyle.getPropertyValue(s), computedStyle.getPropertyPriority(s))
         );
+
+    if(transformDraggedElement) {
```
isaacHagoel, line 80: if u apply the fix i suggested, the 'if' here will not be needed

isaacHagoel (commented):
I tested it. Overall looks good! thanks!
Added a few small comments

src/action.js
isaacHagoel: also, maybe it is a good idea to also pass 'idx' (the index the dragged element will become) into transformDraggedElement... what do u think? useful?

MaddoxDevelopment: I haven't necessarily had a use case for the idx, but it doesn't hurt to send more data just in case. Will update.

README.md
MaddoxDevelopment: Good catch, updated.

src/action.js
MaddoxDevelopment: Updated

src/action.js
MaddoxDevelopment: Updated

src/helpers/styler.js
```diff
@@ -69,10 +71,12 @@ export function morphDraggedElementToBeLike(draggedEl, copyFromEl, currentMouseX
     const computedStyle = window.getComputedStyle(copyFromEl);
     Array.from(computedStyle)
         .filter(s => s.startsWith('background') || s.startsWith('padding') || s.startsWith('font') || s.startsWith('text') || s.startsWith('align') ||
-        s.startsWith('justify') || s.startsWith('display') || s.startsWith('flex') || s.startsWith('border') || s === 'opacity' || s === 'color')
+            s.startsWith('justify') || s.startsWith('display') || s.startsWith('flex') || s.startsWith('border') || s === 'opacity' || s === 'color')
         .forEach(s =>
             draggedEl.style.setProperty(s, computedStyle.getPropertyValue(s), computedStyle.getPropertyPriority(s))
         );
+
+    transformDraggedElement(draggedEl);
```
isaacHagoel, line 79: no need to pass in the draggedEl right? it is a function that takes no params (curried)
`transformDraggedElement()` should be fine

README.md
```diff
@@ -85,17 +85,18 @@ npm install --save-dev svelte-dnd-action
...
+| `dropTargetStyle`         | Object<String> | No                                                           | `{outline: 'rgba(255, 255, 102, 0.7) solid 2px'}` | An object of styles to apply to the dnd-zone when items can be dragged in to it. Note: the styles override any inline styles applied to the dnd-zone. When the styles are removed, any original inline styles will be lost |
+| `transformDraggedElement` | Function       | No                                                           | N/A                                               | A function that is invoked when the draggable element enters the dnd-zone or hover overs a new index in the current dnd-zone. <br />Signature:<br />function(element, data, index) {}<br />**element**: The dragged element. <br />**data**: The data of the item from the items array.<br />**index**: The index the dragged element will become in the new dnd-zone.<br /><br />This allows you to override properties on the dragged element, such as innerHTML to change how it displays. |
```
isaacHagoel, line 96: default value should be `()=>{}` rather than 'N/A' (none required params should have default values)

isaacHagoel (commented):
looks good. added two tiny comments (hopefully will take u 10 seconds to address) and we're good to go

src/helpers/styler.js
MaddoxDevelopment: Woops left that from an earlier test, fixed.

README.md
MaddoxDevelopment: Updated

isaacHagoel:
fixes #106

PR #149, opened 2020-10-14
143 - experiment - improved jsdoc and exported to a type definition file

isaacHagoel:
This is just a quick experiment.
Running `yarn build` generates `index.d.ts` in the dist folder with what seems to be reasonable type definitions. 
I tried to use it within a svelte typescript project (with npm link to a local copy) and it doesn't seem to acknowledge the type definitions or do anything useful, at least with VSCode. Maybe it is because of npm link or the fact that svelte consumes the code directly from src. I am not sure.
Within Intellij, the type definitions seem to be picked up and working properly within .ts files (see image) but proper tooling is missing for .svelte files so I can't check how it behaves in context. 
image

There is a TODO item regarding typing the emitted event (finalize and consider which basically have the same shape)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.iml
 node_modules
 /dist/
+/types/
 cypress/videos
 cypress/screenshots
 
--- a/package.json
+++ b/package.json
@@ -3,20 +3,23 @@
   "svelte": "src/index.js",
   "module": "dist/index.mjs",
   "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "scripts": {
     "test": "cypress run",
-    "build": "rollup -c",
+    "build": "tsc && rollup -c && sed -i '' \"s/^namespace/declare namespace/g\" dist/index.d.ts",
     "prepublishOnly": "npm run build"
   },
   "dependencies": {},
   "devDependencies": {
     "@babel/core": "^7.9.6",
     "@babel/preset-env": "^7.9.6",
+    "@rollup/plugin-node-resolve": "^6.0.0",
     "babel-jest": "^26.0.1",
     "cypress": "^4.5.0",
-    "@rollup/plugin-node-resolve": "^6.0.0",
+    "rollup": "^1.20.0",
     "rollup-plugin-babel": "^4.3.2",
-    "rollup": "^1.20.0"
+    "rollup-plugin-dts": "^1.4.13",
+    "typescript": "^4.0.3"
   },
   "keywords": [
     "svelte",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,4 +1,5 @@
 import resolve from '@rollup/plugin-node-resolve';
+import dts from "rollup-plugin-dts";
 import babel from 'rollup-plugin-babel'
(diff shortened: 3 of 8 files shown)


isaacHagoel:
Actually, with a normal .ts file it works well in VSCode as well. It is only problematic within the .svelte file for some reason.
image

isaacHagoel:
So I continued down the rabbit hole and it turns out custom events are not supported out of the box. 
there is [an open issue for it] (https://github.com/sveltejs/language-tools/issues/431) and a potential workaround (global declaration within the Svelte.jsx namesapce, see issue).  
I used a very silly example to see how vscode + svelte deal with typed actions and encountered this issue.

my silly example App.svelte
```html
<script lang="ts">
	import {act} from './act';
</script>
<section use:act={{a: "5"}} on:test={(e) => {}}>
	Hello
</section>
```

act.ts
```typescript
type Opts = {
  a: string
}
export function act(node: HTMLElement, options: Opts) {
  console.log(node);
  node.dispatchEvent(new CustomEvent('test', { detail: {} }));
  return {
    update: (newoptions: Opts) => { },
    destroy: () => { } 
  }
}
```

isaacHagoel:
More findings:
turns out the vscode plugin has a really strange behaviour. When importing svelte-dnd-action in the root component it can't find the types but when importing from any other component it can.
In order for it not to complain about on:condsider and on:finalise and in order to have type info about the event, I have added this to index.d.ts for `svelte-dnd-action`:
```typescript
declare enum TRIGGERS {
    DRAG_STARTED,
    DRAGGED_ENTERED,
    DRAGGED_OVER_INDEX,
    DRAGGED_LEFT_EVENT_NAME,
    DROPPED_INTO_ZONE,
    DROPPED_INTO_ANOTHER,
    DROPPED_OUTSIDE_OF_ANY,
    DRAG_STOPPED
}
declare enum SOURCES {
    POINTER,
    KEYBOARD
}

type Info = {
    trigger: TRIGGERS;
    id: string;
    source: SOURCES;
};
type DndEvent = {
    detail: {
        items: Array<object>,
        info: Info
    }
}
```
and exported `DndEvent`. 
I couldn't find a way to make JSDoc generate exactly these type definitions automatically (ex: make it create an actual enum, even with the @enum annotation). Looks like if we don't go with full type integration, it is best to create the definitions file manually. 

In the Svelte app I had to add a `global.d.ts` file that looks as follows (will need to be added to the README if we go forward with typescript):
```typscript
declare type DndEvent = { detail: import('svelte-dnd-action').DndEvent };
declare namespace svelte.JSX {
    interface HTMLAttributes<T> {
        onconsider?: function(DndEvent): void
        onfinalize?: (e: DndEvent) => void;
    }
}
```
`tsconfig.json` needs to be edited to include it ` "include": ["src/**/*", "global.d.ts"],` (when i've put it in /src vscode couldn't see it).
Notice: No matter what I tried, within the svelte component it always treats `e` as type `any`, therefore I explicitly added the type to the handler function in the component that consumes the library:
```typescript
function handleSort(e: DndEvent) {
        items = e.detail.items;
}
```
Otherwise, it seems to work normally.

Issue #150, opened 2020-10-16
Dragging the one before last item in a list make the shadow item appear at the end of the list


isaacHagoel:
This happens because the flip animations transforms the element underneath the dragged element so it doesn't understand it is above (need to look at the current position, ignoring any transforms)

PR #151, opened 2020-10-16
150 - ignore transforms when calculating would-be index

isaacHagoel:
fixed the issue. now dragging the one before last element works as expected. added a function that removes the transforms when calculating the would be index


isaacHagoel:
fixes #150

PR #170, opened 2020-10-23
166 - not starting drag when space/enter is pressed on a nested input element

### LINKED ISSUE #166: Dragged content keyboard interactions broken
Hi there, 

Many thanks for this work, very useful for me. I'm trying to use svelte-dnd-action to allow re-ordering of a table rows by the user.
This works fine for all mouse-based operations but breaks when the rows have interactive elements such as a checkbox.
Here is [a REPL](https://svelte.dev/repl/2140934732a14219975549076f11a10c?version=3.29.0).

Each row can be re-ordered fine and has an "Enabled" checkbox which can be clicked alright.

However, if you use the keyboard to focus the checkbox and press Enter or Space, instead of the checkbox being toggled, it's the  svelte-dnd-action wrapper that grabs the focus and a keyboard drag operation is started.

The problem comes from having a global keydown handler that starts the drag operation as soon as Enter/space is pressed. It should only do that if it is the draggable element itself that has focus, not if the focus is on one of its children. 
I have not been able to fix without breaking it all so for now I've just made a copy and added a flag to disable keyboard operation entirely.

Thanks for looking into it.

Best regards

isaacHagoel:
fixes #166

### DIFF
--- a/src/keyboardAction.js
+++ b/src/keyboardAction.js
@@ -159,6 +159,10 @@ export function dndzone(node, options) {
         switch(e.key) {
             case("Enter"):
             case(" "): {
+                // we don't want to affect nested input elements
+                if (e.target.value !== undefined && !allDragTargets.has(e.target)) {
+                    return;
+                }
                 e.preventDefault(); // preventing scrolling on spacebar
                 e.stopPropagation();
                 if (isDragging) {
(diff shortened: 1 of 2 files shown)


src/keyboardAction.js
```diff
@@ -259,16 +263,16 @@ export function dndzone(node, options) {
...
-        node.addEventListener('focus', handleZoneFocus);
-
         if (config.type && newType !== config.type) {
             unregisterDropZone(node, config.type);
         }
         config.type = newType;
         registerDropZone(node, newType);
         dzToConfig.set(node, config);
 
+        node.tabIndex = isDragging && (node === focusedDz || config.dropFromOthersDisabled || (focusedDz && config.type !== dzToConfig.get(focusedDz).type)) ? -1 : 0;
```
isaacHagoel, line 273: not related to this fix. just noticed it is better off in the new place

PR #174, opened 2020-10-26
feat: make debug output configurable

### LINKED ISSUE #165: Make console.debug() output configurable
Internally you're using `console.debug()` to output info to the console. However this seems to be an alias for `console.log()` in most browsers. It would be great to enable/disable the debug output via a boolean config variable.

Is there any other mechanism to disable debug output? Would you be accepting PRs to implement this?

pReya:
closes #165 

* Not sure if you want to have this in `constants.js`, but I thought this is not enough code to create its own file. Let me know what you think about this.

* Also my editor fixed a lot of inconsistent code formatting (indentations, semicolons, single/double quotes, et.). Let me know if this is okay for you, or if you want me to remove it from the PR or move to a second PR for clarity.

* A test for the debug function would be a good idea, but I'm not sure if there's a way to assert console output via Cypress.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -39,8 +39,30 @@ export function overrideItemIdKeyNameBeforeInitialisingDndZones(newKeyName) {
     if (typeof newKeyName !== "string") {
         throw new Error("item id key has to be a string");
     }
-    console.debug("overriding item id key name", newKeyName)
+    printDebug(() => ["overriding item id key name", newKeyName])
     ITEM_ID_KEY = newKeyName;
 }
 
 export const isOnServer =  (typeof window === 'undefined');
+
+export let printDebug = () => {};
+
+/**
+ * Allows the user to show console debug output if he wants
+ * * @param {Boolean} isDebug
+ */
+export function setDebugMode(isDebug) {
+    if (isDebug) {
+        printDebug = (generateMessage) => {
+            const message = generateMessage();
+            if (Array.isArray(message)) {
+                console.debug(...message)
+            } else {
+                console.debug(message)
+            }
+        }
+    }
+    else {
+        printDebug = () => {};
+    }
+}
--- a/src/constants.js
+++ b/src/constants.js
@@ -53,12 +53,12 @@ export let printDebug = () => {};
  */
 export function setDebugMode(isDebug) {
     if (isDebug) {
-        printDebug = (generateMessage) => {
+        printDebug = (generateMessage, logFunction = console.debug) => {
             const message = generateMessage();
             if (Array.isArray(message)) {
-                console.debug(...message)
+                logFunction(...message)
             } else {
-                console.debug(message)
+                logFunction(message)
             }
         }
     }
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,3 @@
 export { dndzone } from './action.js';
 export { alertToScreenReader } from './helpers/aria';
-export { TRIGGERS, SOURCES, SHADOW_ITEM_MARKER_PROPERTY_NAME, overrideItemIdKeyNameBeforeInitialisingDndZones } from './constants';
+export { TRIGGERS, SOURCES, SHADOW_ITEM_MARKER_PROPERTY_NAME, overrideItemIdKeyNameBeforeInitialisingDndZones, setDebugMode } from './constants';
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ It is being actively maintained.
 - Scroll dnd-zones and/or the window horizontally or vertically by placing the dragged element next to the edge
 - Supports advanced use-cases such as various flavours of copy-on-drag and custom drag handles (see examples below)
 - Performant and small footprint (no external dependencies, no fluff code)  
-- Fully accessible (alpha) - keyboard support, aria attributes and assistive instructions for screen readers  
+- Fully accessible (alpha) - keyboard support, aria attributes and assistive instructions for screen readers
 
 ### Installation
 **Pre-requisites**: svelte-3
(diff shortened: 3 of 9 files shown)


src/constants.js
```diff
@@ -41,19 +41,19 @@ export function decrementActiveDropZoneCount() {
...
 }
 
-export const isOnServer = typeof window === "undefined";
+export const isOnServer = typeof window === 'undefined';
 
-export let printDebug = () => { };
+export let printDebug = () => {};
 
 /**
  * Allows the user to show console debug output if he wants
```
isaacHagoel, line 59: Please add an `else` block to allow turning debug mode off after turning it on

src/constants.js
```diff
@@ -34,13 +40,26 @@ export function decrementActiveDropZoneCount() {
...
+            'can only override the id key before initialising any dndzone'
+        );
     }
-    if (typeof newKeyName !== "string") {
-        throw new Error("item id key has to be a string");
+    if (typeof newKeyName !== 'string') {
+        throw new Error('item id key has to be a string');
     }
-    console.debug("overriding item id key name", newKeyName)
+    printDebug(() => 'overriding item id key name', newKeyName);
```
isaacHagoel, line 50: what is going to happen to the second parameter in such cases (`newKeyItem`)?

src/index.js
```diff
@@ -1,3 +1,9 @@
 export { dndzone } from './action.js';
 export { alertToScreenReader } from './helpers/aria';
-export { TRIGGERS, SOURCES, SHADOW_ITEM_MARKER_PROPERTY_NAME, overrideItemIdKeyNameBeforeInitialisingDndZones } from './constants';
+export {
+    TRIGGERS,
+    SOURCES,
+    SHADOW_ITEM_MARKER_PROPERTY_NAME,
+    overrideItemIdKeyNameBeforeInitialisingDndZones,
+    setDebugMode,
```
isaacHagoel, line 8: We should probably mention it in the README file and add a declaration to `types/index.d.ts`

isaacHagoel (changes requested):
1. Overall looks good. Thanks for this. I added a few comments, only one of which might be challenging I think.
2. Regarding testing, maybe in this case we can change printDebug's signature so it looks like this:
```javascript
export function printDebugTo(generateMessage, writerFunction = console.debug) {
   ...
}
```
this will allow the test to pass in a function that just takes the output and puts it in a string or something like that, which is easily testable.
3. Regarding location, I think it's either constants.js or utils.js. constants is probably the better choice.
4. Regarding formatting. I do agree that having inconsistent formatting is not great. On the other hand, I think that most of the automatic changes that were introduced here are not for the better. Especially everything that has to do with line breaks (way too aggressive) and stuff like adding un-needed commas on the last value of an object. I see two options: either we remove it from this PR and do it later as its own thing or we make the formatting rules "better" (whatever that means :)), add the prettier config file to the repo and make it run as a pre-commit hook. Thoughts?

src/index.js
pReya: Done

src/constants.js
pReya: Sorry I did not see these. I changed thes multi-argument outputs to return arrays now. Any other idea?

pReya:
>     2\. Regarding testing, maybe in this case we can change printDebug's signature so it looks like this:

Yeah, this could be one option. Let me read up a little more on Cypress and see if there's another way.

pReya:
> 4. Regarding formatting. I do agree that having inconsistent formatting is not great. On the other hand, I think that most of the automatic changes that were introduced here are not for the better. Especially everything that has to do with line breaks (way too aggressive) and stuff like adding un-needed commas on the last value of an object. I see two options: either we remove it from this PR and do it later as its own thing or we make the formatting rules "better" (whatever that means :)), add the prettier config file to the repo and make it run as a pre-commit hook. Thoughts?

I won't be getting into the old formatting fight ;) I've removed the formatting changes. But it'd be great to have some kind of consistent formatting. Maybe you can add a prettier config according to your preferences in the future.

pReya:
> > ```
> > 2\. Regarding testing, maybe in this case we can change printDebug's signature so it looks like this:
> > ```
> 
> Yeah, this could be one option. Let me read up a little more on Cypress and see if there's another way.

I think it's best to either use your suggestion or use a [Cypress spy](https://docs.cypress.io/api/commands/spy.html#Syntax) on the console object. Any preferences?

isaacHagoel:
@pReya 
Sorry, I just saw your comments now. I am moving house in less than two days so things are a bit chaotic around here. 

> I won't be getting into the old formatting fight ;) I've removed the formatting changes. But it'd be great to have some kind of consistent formatting. Maybe you can add a prettier config according to your preferences in the future.

I agree. I will create an issue for it

>@isaacHagoel I think it's best to either use your suggestion or use a Cypress spy on the console object. Any preferences?

No strong preference.

>Sorry I did not see these. I changed these multi-argument outputs to return arrays now. Any other idea?

I haven't tested it yet but if this solution works, great.

Did you have a chance to look into adding a declaration in `typings/index.d.ts`? should be a 2 minutes thing.

pReya:
@isaacHagoel Would you mind taking another look?

isaacHagoel:
> @isaacHagoel Would you mind taking another look?

@pReya i am moving house today but will have a look (and test it) as soon as i can

isaacHagoel:
@pReya haven't forgotten this. Planning to test and most likely merge tomorrow

pReya:
> @pReya haven't forgotten this. Planning to test and most likely merge tomorrow

No worries, take your time.

isaacHagoel (approved):
Looks and works well. merging to master, doing some super tiny tweaks and releasing. thanks for this!

PR #183, opened 2020-11-02
Drop item by pressing Enter/Space, Cancel with ESC

weeblr:
Hi

Had a first go at this, not sure it matches your standard. It does the job for me but I'm only using it within the context of sorting items inside a single list. The bulk of it is:
- config object has a new `itemsAtDragStart ` prop
- in `handleDragStart`, items are saved into that prop
- in `handleKeyDown`, if ESC is pressed, items are restored to the content of `itemsAtDragStart `
- ArrowUp/Down only trigger `dispatchConsiderEvent`, not finalize
- `handleDrop` triggers `dispatchFinalizeEvent` only when Enter/Space is pressed. And `dispatchConsiderEvent` in all other situations, ie normally when ESC is pressed, cancelling the move. Or if there is a mouse click "outside of anything"

I have left in code I commented out for your reference, let me know.

### DIFF
--- a/src/keyboardAction.js
+++ b/src/keyboardAction.js
@@ -41,9 +41,9 @@ const INSTRUCTION_IDs = initAria();
 function registerDropZone(dropZoneEl, type) {
     printDebug(() => 'registering drop-zone if absent');
     if (typeToDropZones.size === 0) {
-      printDebug(() => "adding global keydown and click handlers");
-      window.addEventListener("keydown", globalKeyDownHandler);
-      window.addEventListener('click', globalClickHandler);
+        printDebug(() => "adding global keydown and click handlers");
+        window.addEventListener("keydown", globalKeyDownHandler);
+        window.addEventListener('click', globalClickHandler);
     }
     if (!typeToDropZones.has(type)) {
         typeToDropZones.set(type, new Set());
@@ -53,6 +53,7 @@ function registerDropZone(dropZoneEl, type) {
         incrementActiveDropZoneCount();
     }
 }
+
 function unregisterDropZone(dropZoneEl, type) {
     printDebug(() => 'unregistering drop-zone');
     typeToDropZones.get(type).delete(dropZoneEl);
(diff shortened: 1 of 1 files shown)


src/keyboardAction.js
```diff
@@ -227,17 +280,23 @@ export function dndzone(node, options) {
                 .filter(dz => dz === focusedDz || !dzToConfig.get(dz).dropFromOthersDisabled),
             dz => dzToConfig.get(dz).dropTargetStyle,
         );
-        dispatchConsiderEvent(node, dzToConfig.get(node).items, {trigger: TRIGGERS.DRAG_STARTED, id: focusedItemId, source: SOURCES.KEYBOARD});
+        config.itemsAtDragStart =  dzToConfig.get(node).items
```
isaacHagoel, line 283: i think there is no need to get it as a config.  you are persisting it on drag start, which is the right thing to do. instead of storing it in the config you can store it as a global like everything else that's related to a drag operation (since there is only one of these at any given time)

src/keyboardAction.js
```diff
@@ -69,19 +70,19 @@ function unregisterDropZone(dropZoneEl, type) {
 
 function globalKeyDownHandler(e) {
     if (!isDragging) return;
-    switch(e.key) {
-        case("Escape"): {
-            handleDrop();
-            break;
-        }
-    }
+    // switch (e.key) {
```
isaacHagoel, line 73: no reverting?

isaacHagoel:
@weeblr thanks. I had a brief look. added two questions. haven't played with it but I think the change will need to go deeper in order to work across the board. It might help if we first make a short spec of what needs to be dispatched/ happen with every action:
- entering dragging mode (space/enter while on an item)
- pressing arrow keys
- tabbing into another list
- hitting space/enter while in another list (drop)
- hitting escape anywhere (cancel)
- clicking outside of everything (cancel?)
- clicking on another item anywhere (we start dragging it, what happens to the item we were dragging before)

isaacHagoel:
@weeblr another thing. i added prettier config to the project. please pull from master. thx.

PR #185, opened 2020-11-09
Include Safari's webkitUserSelect

alexobenauer:
This fixes some bugs I've experienced while dragging items in Safari.

Edit: I should have mentioned that Safari needing the "webkit" prefix on this property in Javascript is [documented here](https://www.w3schools.com/jsref/prop_style_userselect.asp).


isaacHagoel:
thanks for this. merging and releasing

PR #203, opened 2020-12-23
Fix for quotes position

chbert:


isaacHagoel:
@chbert  thanks for this

PR #222, opened 2021-01-13
Fix typo in Typescript instruction snippet

Azza292:
Fix PR fixes a typo that will cause errors when copying the Typescript code snippet.


isaacHagoel:
@Azza292 thanks for this

Issue #223, opened 2021-01-13
Event for getting current position


Florian-Schoenherr:
I have implemented some dnd stuff myself and found that I really wish for a way to retrieve the current position while dragging, so that I can potentially reconfigure the dragged element to be `position: absolute;`. I know you're using translate3d, but that would work the same way. Maybe an event on _every_ movement would be overkill.
I also thought about a separate dropzone for whiteboard-like absolute positioned dnd and then just changing the drag-elements type when dragging it over a different dropzone. Then just react to data changes.
🤔

isaacHagoel:
Hi @Florian-Schoenherr,
Not sure i fully follow. Is this a feature request/ a question/ something else. 
any of the above is fine, i just need some help so i can respond properly.

all of the events emitted by the library include the id of the dragged item (under info) which can be used to retrieve the index very easily. alternatively, it can be added to the info very easily if needed. no one has asked for it so far.
Internally the library tracks the "would be index" during drag operations in order to decide when to emit which event.

regarding the whiteboard. i will need some more info or an example in order to respond.

Florian-Schoenherr:
You probably know apps like diagrams.net/excalidraw/miro.com (although they differ in implementation). I want to use something like that, but also want to use your "sortable-list"-type dnd.
Let's say I have 3 cards, which I drag around with absolute positioning (like miro, ..).
Now I want to drag one card onto another and that's where your dndzone should kick in. If I then decide to drag one card out of your dndzone (which has 2 cards) and just "onto the whiteboard", it should be positioned absolutely again.

That would mean we'd need something like dropinside, dropoutside and the current position of the dropped element. If that's out of scope, it's okay.

"Whiteboard" would then mean, maybe we want a completely separate "dnd-absolute" for absolutely positioned elements and then people would be able to drag&drop between your dnd and "dnd-absolute"? Maybe I'm overthinking.

isaacHagoel:
this type of "whiteboard" drop zone makes different assumptions than a sortable dropzone.
For example, a sortable dropzones assumes elements can't overlap (the centre of the dragged element can never be over two elements at the same time). In a white board they can, which brings the stacking order into the equation (which element should be drawn on top of which etc.). 
Keyboard interactions might also be different, as well as aria labels and messages.
This mix might be possible though with some tinkering, assuming your app implements the whiteboard part. The tricky part is of course crossing the border of the whiteboard in and out (each one requires a different handover)

[...]

PR #227, opened 2021-01-15
Add 'customDrop' config function

canadaduane:
Allows for more complex interactions between svelte-dnd-action and non-svelte-dnd-action drop points.

For example, use `customDrop` function to determine if a custom element should be used as a drop point such as a "Trash Bin" to delete an item from a sortable list.

- accepts dragged element data and mouse position as params
- return true or undefined to complete revert-to-origin animation (false will cancel the animation).

### DIFF
--- a/src/action.js
+++ b/src/action.js
@@ -50,6 +50,7 @@ function validateOptions(options) {
         dropTargetStyle,
         dropTargetClasses,
         transformDraggedElement,
+        customDrop,
         autoAriaDisabled,
         ...rest
     } = options;
--- a/src/pointerAction.js
+++ b/src/pointerAction.js
@@ -141,7 +141,8 @@ function handleDraggedEntered(e) {
         dispatchConsiderEvent(originDropZone, newOriginZoneItems, {
             trigger: TRIGGERS.DRAGGED_ENTERED_ANOTHER,
             id: draggedElData[ITEM_ID_KEY],
-            source: SOURCES.POINTER
+            source: SOURCES.POINTER,
+            target: dzToConfig.get(e.currentTarget)
         });
     } else {
         const shadowPlaceHolderIdx = findShadowPlaceHolderIdx(items);
@@ -256,7 +257,13 @@ function handleDrop() {
         unDecorateShadowElement(shadowElDropZone.children[shadowElIdx]);
         cleanupPostDrop();
     }
-    animateDraggedToFinalPosition(shadowElIdx, finalizeWithinZone);
+
+    const showAnimation = dzToConfig.get(originDropZone).customDrop(draggedElData, currentMousePosition);
+    if (showAnimation === undefined || showAnimation === true) {
+        animateDraggedToFinalPosition(shadowElIdx, finalizeWithinZone);
+    } else {
+        finalizeWithinZone();
+    }
 }
 
 // helper function for handleDrop
@@ -302,7 +309,8 @@ export function dndzone(node, options) {
         dropFromOthersDisabled: false,
         dropTargetStyle: DEFAULT_DROP_TARGET_STYLE,
         dropTargetClasses: [],
-        transformDraggedElement: () => {}
+        transformDraggedElement: () => {},
+        customDrop: () => {}
     };
     printDebug(() => [`dndzone good to go options: ${toString(options)}, config: ${toString(config)}`, {node}]);
     let elToIdx = new Map();
@@ -420,7 +428,8 @@ export function dndzone(node, options) {
         dropFromOthersDisabled = false,
         dropTargetStyle = DEFAULT_DROP_TARGET_STYLE,
         dropTargetClasses = [],
-        transformDraggedElement = () => {}
+        transformDraggedElement = () => {},
+        customDrop = () => {}
     }) {
         config.dropAnimationDurationMs = dropAnimationDurationMs;
         if (config.type && newType !== config.type) {
@@ -432,6 +441,7 @@ export function dndzone(node, options) {
         config.items = [...items];
         config.dragDisabled = dragDisabled;
         config.transformDraggedElement = transformDraggedElement;
+        config.customDrop = customDrop;
 
         // realtime update for dropTargetStyle
         if (
--- a/src/pointerAction.js
+++ b/src/pointerAction.js
@@ -141,8 +141,7 @@ function handleDraggedEntered(e) {
         dispatchConsiderEvent(originDropZone, newOriginZoneItems, {
             trigger: TRIGGERS.DRAGGED_ENTERED_ANOTHER,
             id: draggedElData[ITEM_ID_KEY],
-            source: SOURCES.POINTER,
-            target: dzToConfig.get(e.currentTarget)
+            source: SOURCES.POINTER
         });
     } else {
         const shadowPlaceHolderIdx = findShadowPlaceHolderIdx(items);


isaacHagoel:
Hi,
Thanks for this. highly appreciated!
1. Can you please explain how this would be used in more detail and ideally also share a usage example (as in code that uses this new option)?
2. if it is expecting a function maybe the name could start with a verb (doBlaOnDrop)
3. I wonder if a cancellable event ('itemDropped' or a similar name) or something like that would be more appropriate, but that depends your answer to \#1
4. If we go ahead with this as is, we will need to update the readme, typings, version (0.9.0) and release-notes

canadaduane:
Is there a way to use the svelte REPL to show this in action, before it's included in a release?

isaacHagoel:
I don't think so, but you can paste some code here as a
comment/attachment/link and i will switch to your branch and try it out
locally

canadaduane:
Here is an example:

https://github.com/canadaduane/svelte-dnd-action-custom-drop

canadaduane:
Screencast:

https://user-images.githubusercontent.com/129/104797103-dbc8e400-5778-11eb-9b53-f6f27acf1010.mp4

isaacHagoel:
From a brief look at the code and mp4 a quick question: can't it be achieved using the existing features?
The finalise handler knows which dropzone the item was dropped to (e.currentTarget) which is what u r using the mouse ciordinates from.
Turning the animation off should be possible by setting flipDuration to zero (can also be done dynamically if needed). 
What am i missing?

isaacHagoel:
(cont): i mean also making the droptargets dnd zones that simply don't display the items.
if my description above is too hand-wavy, i can try to create it in the REPL when I have 30-45 min.
still, if u do understand my hand waves and noticed something i missed, pls do let me know and save me from wasting my time on an impossible attempt 😄

canadaduane:
I tried with existing features for a very long time. It might be possible,
but it was very difficult to achieve if so.

There are multiple outcomes at a high level that get difficult to reason
about inside of a "TRIGGERS" switch inside consider/finalize:
- I don't want the dropped items to go into the dropzone itself, I want
them to go to the sortable list associated with the dropzone
- Several times, I ran into issues where spacers (hidden shadow elements)
would exist where I didn't want them to
- I couldn't use "Dragula" style example, because I also need the user to
be able to re-arrange the order inside the list
- The COPY/MOVE/DELETE actions are easier to reason about on drop

The biggest issue for me was spending about a week trying to figure out the
internals of svelte-dnd-action using a TRIGGERS switch, and failing at it.
There were too many edge cases, and I found this was a much more elegant
solution for me.

It's possible setting flipDuration to zero may be sufficient for turning
the animation off--I hadn't considered that.

isaacHagoel:
I think this usecase is super valid and i'd like the library to support it. I will look at your code in more depth and also try to do it using existing features to get a feel for what makes it complicated. I am sorry to hear you had a hellish week fighting with the library. Feel free to reach out earlier if that happens again.
 Will keep you posted.

canadaduane:
Sounds good, and thank you. I appreciate the library you all have made
here--I tried other non-svelte library solutions (e.g. SortableJS, Dragula)
and they, too, were quite complicated to get right for this use case.

isaacHagoel:
Hi,
I've made this POC [REPL](https://svelte.dev/repl/193b40b412204176831d87358e02d625?version=3.31.2).
Make sure the REPL downloads v0.8.6 (there was a bug in 0.8.5 that will affect it). A hard refresh should get you the correct version.

There are some things that aren't perfect about it (tabindex on the delete group, the way the element transforms and the lack of styling) but I think it functions as expected. Have I missed anything?
Pls have a look.

canadaduane:
This looks really good. The only thing that's missing is a "Favorites" folder that behaves differently than the others, i.e. it results in accepting a copy of a file rather than the file itself. Some constraints:

1. If you drag the same file multiple times into Favorites, only one copy will ever be made (you can't "favorite" a file twice; another way of saying this is that Favorites is an Ordered Set, not a List)
2. If you drag a favorite into a Folder, a copy is made (i.e. the favorite stays in Favorites)
3. If you drag a favorite into a Folder and the original file is already in that Folder, no copy is made (no action).

isaacHagoel:
Sounds a tiny bit strange to be able to drag from favorites to a folder. I
would expect to only be able to delete a favourite (dragging to trash).

With that said, it sounds very doable either way. Up to you :)

It will require treating the names as unique within folders (which makes
sense).
I am happy to add it but will it be okay to ask you to style the resulting
repl a bit so that it looks nicer (like your app)? That way we will be able
to include it in the readme as a super advanced example.

canadaduane:
Sure, I can style it, no problem.

Yeah, the "Files" and "Folders" metaphor breaks down a bit here. How we're
actually using this is as a set of collections of images in our app (
staging.relm.us). Certain admins have the ability to add an image to any
public collection from their private "favorites".

canadaduane:
Should I wait until you're done before styling? I'm not sure if the REPL
has the ability to merge/rebase/resolve conflicts.

isaacHagoel:
Cool. I am not next to a computer atm.
Will have a look and add the favorites today or tomorrow

isaacHagoel:
After im done. Thx

isaacHagoel:
Alright. This was more straightforward than I expected (not complaining) and overall the library seems to have lended itself quite well to this usecase imho.
[REPL](https://svelte.dev/repl/443590a8ce1f459fbb790cdf6fd6711a?version=3.31.2)

Things that are still missing:
- Styling
- Applying `transformDraggedElement` in order to make the morphing effect nicer (should it more to some icon or something?)
- It is not good that the delete button has a `tabindex` of 0 but currently the library auto-applies it. I will think about a proper solution for this at the library level (open for ideas).
- keyboard interactions seem to work (I tested in a rush) but it doesn't feel great (it's not clear what's happening) and the aria messages and instructions need to be overridden to something that makes sense (using `autoAriaDisabled` on the dnd-zones ).

canadaduane:
Nice work!!

I've updated styles here: https://svelte.dev/repl/317cf511ee4445b8b36e2ff13e2327c4?version=3.31.2

The only quibble I have is that I'd prefer it if the drag didn't make it look like it will MOVE the item by default. In my example above, it defaults to making a copy (leaving the shadow element in place) so that whatever the user does, the item doesn't look like it's moving until the very last moment when it's dropped into a folder. This allows dragging to the favorites (for example) to look like a copy is being made during the drag. Not sure if this is a big change though.

isaacHagoel:
It should be easy to make it always looks like either move or copy. The thing here is that it is sometimes move sometomes copy and sometimes delete...

isaacHagoel:
After giving this some more thought and in the context of the recent/ on-going discussion around #223, I have kind of gone a full circle.
I am now considering dispatching a new `consider` event that gets fired on the origin dndzone when the item is dropped  anywhere outside of the origin dndzone, before the final animation to place plays (== before the `finalize` events).
The event can include (in e.detail) information such as the current mouse/touch position, the dragged dom element (which will allow transforming the element or calculating the distance between the mouse and the centre/edges of the element for example).
If the event is canceled (e.preventDefault()), the final animation won't play (and maybe the last `finalize` events won't fire, not sure about that one).

this will allow not using dndzones for `favourites` in this example (i think it is still a good idea to make the other folders and delete buttons dndzones since "move" semantics is appropriate for them ), but still be able to respond when something was dropped onto favourites.
It will probably also be handy in the whiteboard example and will prevent the application code from having to track the mouse on its own. it will also prevent the jitter on drop I think.
This idea is very close to the original feature you proposed in this PR, just uses a cancellable event instead of a new option..
I haven't thought it through yet, but this is what's on my mind at them moment. Keyboard might present a challenge here as well and will require consideration.
Thoughts?

canadaduane:
I think that makes sense. There's a lot that svelte-dnd-action does automatically for sortable lists, but then there are some clear dnd use cases that are slightly out of scope (favorites, whiteboard) and it would be nice to be able to get just enough information at the application level to implement alternatives.

canadaduane:
One more thought: An additional requirement that I will have is dragging an item (image) into the 3D world (a canvas) and then turning it into a 3D object inside the world. I think what you propose would help with that as well.

Florian-Schoenherr:
I mean, if we get the nodes data (with mouse position), essentially all positioning is opened up to the dev. That's really good!
It should solve the whiteboard stuff.

isaacHagoel:
@canadaduane @Florian-Schoenherr 
I created a draft PR, #232. I tried to add the minimum that I thought would be required. I included an example in the PR's description.
I didn't have time to try to use it in our two use cases just yet, but in case you want to play you can try the branch (223-user-dropped-event) locally.

PR #232, opened 2021-01-19
223 - added a cancelable USER_DROPPED consider event on the origin zone

### LINKED ISSUE #223: Event for getting current position

isaacHagoel:
added a cancelable USER_DROPPED consider event on the origin zone if the dragged element was dropped outside of any. cancelling skips the final animation to place and dispatches the finalize events right away.

missing updates to README and types

resolves #223

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -7,6 +7,7 @@ export const TRIGGERS = {
     DRAGGED_OVER_INDEX: DRAGGED_OVER_INDEX_EVENT_NAME,
     DRAGGED_LEFT: DRAGGED_LEFT_EVENT_NAME,
     DRAGGED_LEFT_ALL: "draggedLeftAll",
+    USER_DROPPED: "userDropped",
     DROPPED_INTO_ZONE: "droppedIntoZone",
     DROPPED_INTO_ANOTHER: "droppedIntoAnother",
     DROPPED_OUTSIDE_OF_ANY: "droppedOutsideOfAny",
--- a/src/helpers/dispatcher.js
+++ b/src/helpers/dispatcher.js
@@ -26,7 +26,7 @@ export function dispatchFinalizeEvent(el, items, info) {
  * @param {Info} info
  */
 export function dispatchConsiderEvent(el, items, info) {
-    el.dispatchEvent(
+    return el.dispatchEvent(
         new CustomEvent(CONSIDER_EVENT_NAME, {
             detail: {items, info}
         })
--- a/src/pointerAction.js
+++ b/src/pointerAction.js
@@ -261,7 +261,21 @@ function handleDrop() {
         unDecorateShadowElement(shadowElDropZone.children[shadowElIdx]);
         cleanupPostDrop();
     }
-    animateDraggedToFinalPosition(shadowElIdx, finalizeWithinZone);
+    let shouldFinalizeDrop = true;
+    if (isDraggedOutsideOfAnyDz) {
+        shouldFinalizeDrop = !dispatchConsiderEvent(originDropZone, dzToConfig.get(originDropZone).items, {
+            trigger: TRIGGERS.USER_DROPPED,
+            id: draggedElData[ITEM_ID_KEY],
+            source: SOURCES.POINTER
+        });
+    }
+    if (shouldFinalizeDrop) {
+        printDebug(() => "will animate before finalizing drop");
+        animateDraggedToFinalPosition(shadowElIdx, finalizeWithinZone);
+    } else {
+        finalizeWithinZone();
+        printDebug(() => "dropped outside and event was cancelled, finalizing without animating");
+    }
 }
 
 // helper function for handleDrop


canadaduane:
This looks good. I will try to use in my dnd implementation in Relm and see if anything comes up.

isaacHagoel:
let me know if you have any feedback. I am hoping to spend more time on this over the weekend

canadaduane:
I will have time tomorrow (Friday) to visit this. Apologies for the delay.

canadaduane:
All my tests check out! This accomplishes what my 'customDrop' PR does and without introducing a new config option.

isaacHagoel:
mmm... i've been playing with it but on my new example (with files and folders) it looks like for some reason the event is always cancelled (without me calling prevent default). I am not sure why. still investigating.

isaacHagoel:
alright. i fixed it. now it behaves correctly in both cases. will proceed testing it in context later.

Florian-Schoenherr:
Sorry, I was blocked this week, can take a look later. If this helps his usecase, you can merge this at any time.

canadaduane:
Is this ready to be merged in?

isaacHagoel:
sorry. I found some issues while testing it.
also, I started to doubt this approach in general, doubting it can go the
whole way.
does it work perfectly for you?

canadaduane:
Yes, it seems to work well. It was a near drop-in replacement for the customDrop approach I'd tried earlier (no pun intended!)

isaacHagoel:
Alright. I will resume looking into it

canadaduane:
Thank you!

isaacHagoel:
@canadaduane can you remind me pls... is there a reason i decided to dispatch the new "USER_DROPPED" event only when the dragged element is outside of any zone? why didn't i make the library dispatch it and allow cancelling the final animation regardless of where it was dropped (which makes more sense to me when i think about it now)?

also, note to self - instead of conditional logic around whether or not to play the animation we can set the duration of the animation to zero if the event was cancelled

isaacHagoel:
reply to self: I did it this way because otherwise the user code doesn't have enough information to decide whether to cancel the drop or not (dropped where). 
I streamlined the implementation and now it doesn't seem to cause any undesired side effects.
using it with the modified code:
```javascript
const {items: newItems, info: {trigger, pointerClientXY, isDroppedOutsideOfAnyDzOfType}} = e.detail;
        if (trigger === TRIGGERS.USER_DROPPED && isDroppedOutsideOfAnyDzOfType) {
            e.preventDefault();
        }
```
Now I wonder whether this is the time and place to introduce a new event type in addition to `consider` and `finalize`. Maybe it should be called `fyi` or something like that. Otherwise this will cause everyone to needlessly re-render when they get this `consider`. not sure if it is a big deal.
If i do add the event, the DRAGGED _STOPPED keyboard event probably falls into this category too (breaking change - version will become 1.0.0).
Any thoughts are welcome.

Also, @Florian-Schoenherr did u get to try it with the whiteboard example?

Florian-Schoenherr:
I'm gonna try it now.

isaacHagoel:
@Florian-Schoenherr sorry, i only saw your edit now. you use 'npm link' in order to make your repo use the local version of the library.
cd to the svelte-dnd-action folder and run `npm link`
cd to your project folder and run `npm link svelte-dnd-action`

Florian-Schoenherr:
image
The white background is a dnd-zone, holding the green cards, which are absolute positioned (with `TRIGGERS.USER_DROPPED`).
The pinkish rectangles inside the cards work like before (you can sort them and you can drag them outside to put them into their own card).

image
Inside `handleDrop()`, I added this for debugging, mainly because I get an error when I drop the last element of the card-list, shortly after this line.

image
Here, `shadowElIdx: 2` is the index of the last element in the card-list, so `getBoundingRectNoTransforms(shadowElDropZone.children[shadowElIdx]);` will not work, because the child with index `2` is currently getting dragged.

isaacHagoel:
not sure i follow. it might be an issue that was recently fixed on the master branch. can u pls provide some files with code that i can used in order to reproduce this error?

isaacHagoel:
@Florian-Schoenherr ^

Florian-Schoenherr:
`getBoundingRectNoTransforms(shadowElDropZone.children[shadowElIdx]);`
in this case will do
`getBoundingRectNoTransforms(shadowElDropZone.children[2]);`
but 
`shadowElDropZone.children` is `[0: div, 1: div]`, so does not have an element at index 2.

But I found out what the bug was:
In one of the `consider` functions I did `itemGroups = itemGroups` instead of `itemGroups = e.detail.items;` 😅 

Now after that change, I can do:
```js
const considerGroup = (e) => {
	if (e.detail.info.trigger === TRIGGERS.USER_DROPPED) {
		const { x, y } = e.detail.info.pointerClientXY;
		const group = itemGroups.find(g => g.id === "id:dnd-shadow-placeholder-0000");
		group.x = x;
		group.y = y;
           	e.preventDefault();
	}
	itemGroups = e.detail.items;
}
```
but finalize still runs, so uhh, would I then replace the shadow item myself?
If I leave away finalize completely and instead of the above, change it to:
```js
group.x = x;
group.y = y;
group.id = e.detail.info.id;
delete group["isDndShadowItem"];
e.preventDefault();
```
it works.

`SHADOW_PLACEHOLDER_ITEM_ID` doesn't seem to be exported any longer

isaacHagoel:
it is an older branch that's why some newer features might not be available.
do i need to check/debug something or does it meet your requirements (this
branch)?

Florian-Schoenherr:
@isaacHagoel do you think the code in the REPL makes sense? I just hacked that together somehow, seemed a little like I was fighting the intended way, otherwise yeah, it works.

isaacHagoel:
I haven't really looked (i am on vacation this weekend). Will let you know

Florian-Schoenherr:
@isaacHagoel have a nice weekend :)

isaacHagoel:
what is the desired behaviour in that REPL? freely dragging the groups around and be able to move items between groups?

Florian-Schoenherr:
Yeah, dragging the groups around, moving between groups and when dragging one item outside a group, adding it to it's own, new group. It works but seems hacky.

isaacHagoel:
It does seem hacky. Will try to find time to improve it tomorrow

isaacHagoel:
I played with it a bit and indeed this usecase (especially the need to create a new group when the item is dropped outside of any, remove the item manually from the original group and so on) doesn't feel nice to implement.
I pushed a few updates to the branch:
1. merged master
2. added `draggedItemData` to the new event's info. This allows you to place it where it needs to go (ex: within a newly allocated group) without worrying about shadow item shenanigans (it is the original item's data). 
As i said, tbh i am still not sure on whether the library should support this kind of usecase or not. If the answer is yes it shoud, i think it should do it in a nicer way.

Florian-Schoenherr:
@isaacHagoel will take a look soon.
I wonder if there's a way to keep all this dnd/whiteboard/Shopify-draggable usecase stuff idiomatic, easy to compose etc.
Of course you can't expand scope endlessly (although I think you've done a great job with the current functionality and would probably be able to do more of this kind of thing).

canadaduane:
Yes, composability & ability to think about these actions / outcomes
separately is a very nice-to-have. I found it slightly difficult to think
about internal state of svelte-dnd-action while also crafting the UI
concepts I needed for my app. I'm grateful for your work here, though, as I
tried several other libraries and it wasn't "easy" in those cases either.

PR #242, opened 2021-02-05
Allow child buttons to be pressed with keyboard

danny-andrews:
**Before**

Notice, how, even though the button is focused, I can't press it with spacebar. Doing so initiates drag mode.

before

**After**

After the change, I can press fire buttons with spacebar. And I can also initiate drag mode by selecting the containing element.

after

### DIFF
--- a/src/keyboardAction.js
+++ b/src/keyboardAction.js
@@ -170,7 +170,7 @@ export function dndzone(node, options) {
             case "Enter":
             case " ": {
                 // we don't want to affect nested input elements or clickable elements
-                if ((e.target.click !== undefined || e.target.value !== undefined || e.target.isContentEditable) && !allDragTargets.has(e.target)) {
+                if ((e.target.disabled !== undefined || e.target.href || e.target.isContentEditable) && !allDragTargets.has(e.target)) {
                     return;
                 }
                 e.preventDefault(); // preventing scrolling on spacebar
--- a/src/keyboardAction.js
+++ b/src/keyboardAction.js
@@ -169,8 +169,8 @@ export function dndzone(node, options) {
         switch (e.key) {
             case "Enter":
             case " ": {
-                // we don't want to affect nested input elements
-                if ((e.target.value !== undefined || e.target.isContentEditable) && !allDragTargets.has(e.target)) {
+                // we don't want to affect nested input elements or clickable elements
+                if ((e.target.click !== undefined || e.target.value !== undefined || e.target.isContentEditable) && !allDragTargets.has(e.target)) {
                     return;
                 }
                 e.preventDefault(); // preventing scrolling on spacebar


src/keyboardAction.js
```diff
@@ -169,8 +169,8 @@ export function dndzone(node, options) {
         switch (e.key) {
             case "Enter":
             case " ": {
-                // we don't want to affect nested input elements
-                if ((e.target.value !== undefined || e.target.isContentEditable) && !allDragTargets.has(e.target)) {
+                // we don't want to affect nested input elements or clickable elements
+                if ((e.target.click !== undefined || e.target.value !== undefined || e.target.isContentEditable) && !allDragTargets.has(e.target)) {
```
isaacHagoel, line 173: good catch. i am not sure that checking for .click is a reliable enough test. looks like a simple div has that too.
Screen Shot 2021-02-05 at 2 56 16 pm

danny-andrews: Yeah, you're right about that. I meant to add a note to it in my PR saying that I wasn't sure about that heuristic. We could of course check for certain tags, like `a` or `button` but that won't work for custom elements.

It makes me think we should only activate dragging when the root drag element is selected. Intercepting keyboard input events for all nested elements and preventing them from bubbling just seems like a recipe for disaster. Maybe we should remove all specific checks and just ensure the clicked element is a drag target:

```js
if ( !allDragTargets.has(e.target)) {
```

Since drag targets are given a tabIndex, they will always be focusable, and keyboard users will be able to move stuff around as needed.

What do you think?

isaacHagoel:
I think: what about drag handles? :)

danny-andrews:
Hmmmm...well, based on the example for drag handles, you already have to define your own onkeydown listener: https://svelte.dev/repl/4949485c5a8f46e7bdbeb73ed565a9c7?version=3.24.1.

isaacHagoel:
It's not the one that initiates the drag it only allows it

danny-andrews:
I don't follow.

danny-andrews:
It initiates the drag by setting `dragDisabled = false`, right?

isaacHagoel:
No. Setting drag disabled to false simply allows the drag to start as it
normally would. It does not trigger the drag.
When drag disabled is true it is not possible to drag the item when it is
false it becomes possible (see readme).

danny-andrews:
I see. What about adding a config option, `dragHandleSelector`? That way, we could look to see if `el.target` and see if it matched that selector. If it did, we would do our drag logic like usual. Only problem with this is that it would be a breaking change, unless it defaulted to '*' or something.

isaacHagoel:
Yeah. A breaking change should be a last resort and super worth it.
I think you original solution was good, it just needs a better heuristic.
Maybe we can introduce a boolean data property like `data-dnd-ignore` and
any element that has it will not initiate drag.
It gives the user full control and is not breaking.

danny-andrews:
Alright, I refined the heuristic to

```js
if ((e.target.disabled !== undefined || e.target.href || e.target.isContentEditable) && !allDragTargets.has(e.target))
```

All buttons and inputs have a `disabled` property (including any sensible custom elements) and all links have an `href` property.

isaacHagoel:
looks promising. i will test it today or tomorrow and if all good will merge. thx

PR #244, opened 2021-02-06
Use touch coords

geovie:
Hi, first thanks for this awesome library.

When using it on mobile (touch devices), especially with bigger drag elements it can be difficult to put the center of the drag element inside a drop zone.

This PR allows to switch between the default behavior (of using the center of the dragged element) and using the mouse/touch coordinate to check the underlying drop zone.

The first commit adjust the code a bit so that both cases can be handled by the code.   
The second commit adds the new behavior and config options.

### DIFF
--- a/src/helpers/observer.js
+++ b/src/helpers/observer.js
@@ -57,7 +57,7 @@ export function observe(draggedEl, dropZones, intervalMs = INTERVAL_MS) {
         // this is a simple algorithm, potential improvement: first look at lastDropZoneFound
         let isDraggedInADropZone = false;
         for (const dz of dropZonesFromDeepToShallow) {
-            const indexObj = findWouldBeIndex(draggedEl, dz);
+            const indexObj = findWouldBeIndex(currentCenterOfDragged, dz);
             if (indexObj === null) {
                 // it is not inside
                 continue;
--- a/src/helpers/observer.js
+++ b/src/helpers/observer.js
@@ -16,13 +16,23 @@ const TOLERANCE_PX = 10;
 const {scrollIfNeeded, resetScrolling} = makeScroller();
 let next;
 
+let mouseX = 0;
+let mouseY = 0;
+
+function onMouseMove(event) {
+    mouseX = event.touches ? event.touches[0].clientX : event.clientX;
+    mouseY = event.touches ? event.touches[0].clientY : event.clientY;
+}
+
 /**
  * Tracks the dragged elements and performs the side effects when it is dragged over a drop zone (basically dispatching custom-events scrolling)
  * @param {Set<HTMLElement>} dropZones
  * @param {HTMLElement} draggedEl
  * @param {number} [intervalMs = INTERVAL_MS]
  */
-export function observe(draggedEl, dropZones, intervalMs = INTERVAL_MS) {
+export function observe(draggedEl, dropZones, intervalMs = INTERVAL_MS, useTouchPosition = false) {
+    document.body.addEventListener("mousemove", onMouseMove);
+    document.body.addEventListener("touchmove", onMouseMove);
     // initialization
     let lastDropZoneFound;
     let lastIndexFound;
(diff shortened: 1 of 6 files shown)


src/pointerAction.js
```diff
@@ -98,7 +98,11 @@ function watchDraggedElement() {
         MIN_OBSERVATION_INTERVAL_MS,
         ...Array.from(dropZones.keys()).map(dz => dzToConfig.get(dz).dropAnimationDurationMs)
     );
-    observe(draggedEl, dropZones, observationIntervalMs * 1.07);
+    const useTouchPosition = Array.from(dropZones.keys())
```
isaacHagoel, line 101: looks like using this option for one dndzone will apply to all. this could be kind of problematic because maybe for some it would not be ideal

src/helpers/observer.js
```diff
@@ -16,13 +16,23 @@ const TOLERANCE_PX = 10;
 const {scrollIfNeeded, resetScrolling} = makeScroller();
 let next;
 
+let mouseX = 0;
+let mouseY = 0;
+
+function onMouseMove(event) {
```
isaacHagoel, line 22: pointerAction already listens on mousemove/touchmove and has the info. i wonder if it could be leveraged.

isaacHagoel (commented):
First and foremost, thanks a lot for this. I appreciate it and think you identified a real issue.

I only added 2 comments that are pretty important I think. There are other small stuff i could comment on (need to add the new option to the main action, types, readme and things like that). 
I stopped commenting because I think we need to take a step back and look at the more fundamental question here: in which situations would the user of the library want to use this option? is there a way for the library to detect this situations automatically and "just work"?
I guess a first step in answering is to enumerate some of this scenarios/ relationships... does it have to do with the ratio between the dragged element and X (where X could be   screensize, dropzone size, element underneath size or whatever)

geovie:
Actually this should've been more of a draft to start the discussion.

I modified a REPL [here](https://svelte.dev/repl/deb90f97bee94b92ae5d52590cc80ce0?version=3.31.2) to show my use case for this.

The first problem is that due to the dragged element changing it's size (and therefore also it's center) different zones will become 'considered' in a loop:

https://user-images.githubusercontent.com/29738679/107142836-cd15bd00-6931-11eb-869a-d6b037b9a19c.mov

This can be solved by using the mouse position as point to detect if a drop zone should be considered.

The second issue is when the item to be dragged is way longer then in some cases it can be nearly impossible to drop the item in certain zones:

https://user-images.githubusercontent.com/29738679/107143001-dd7a6780-6932-11eb-866f-22b45fedefb5.mov

As for a way to detect this IMO enabling this on touch devices by default might be a good solution.    
I'm not sure if there is a good solution with detecting (and enabling) this based on the ratio of drag/drop element....

isaacHagoel:
I will give it more thought but i wonder... Why is there any difference in
your mind between touch and mouse for the examples above?

geovie:
I can't say exactly the difference, but I guess with touch it's more intuitive that the drop 'happens' where the finger is moved to and also I think on mobile there is less space available so this case can happen more often.     

So it should still be configurable, but maybe it would make sense to default to it on mobile.

isaacHagoel:
i thought about it some more, mainly around how to not fall back into the very issues that using the centre of the element rather than the pointer was addressing in the first place. i have an idea you might find interesting:  
what if we add a way (option) to position the drag element such that its centre is right on the mouse pointer when drag starts (it can simply "jump" into this position when taken out of the original list)?
that way, both become one and the same and both of the problems you illustrated above should be resolved.

geovie:
I like the proposal, would it be for all drop zones or could it be configured for each drop zone?

E.g. in the case of the item drop zone in the REPL it could be confusing if the items 'jump' around when re-ordering the items.

So item drop zone without jump and collection drop zone with jump to center, and I'd be okay if the drag point would stay in the center once the drag element is moved back to the item drop zone.

isaacHagoel:
I think it should happen on dragStart and not change again throughout the
duration of the drag.
That means that when configured on a dndzone, it affects drag operations on
items that are taken out of the zone (so it is not global in that sense).
Is that powerful enough?

isaacHagoel:
Hi @geovie,
i've implemented the idea i described above (dragged element positions its center on cursor on drag start) in this branch: `249-dragged-jump-to-cursor-on-start`. currently it is not optional (i still need to add that part).
Can you give it a try?
see also #250

geovie:
Thanks @isaacHagoel, this works for the drop zone, a bit annoying is the move when you use a drag handle (e.g. https://svelte.dev/repl/4949485c5a8f46e7bdbeb73ed565a9c7?version=3.24.1).   
Using the drag handle center in this case is probably to much of a special case or?

isaacHagoel:
This just a demo so it's always "on".
If we go with it it will be optional and "off" by default. Yeah, it makes
no sense to turn it on for zones with drag handles.

isaacHagoel:
I merged #250. I think it addresses this issue quite well (shouldn't be turned on when using drag handles).
closing this for now.

Issue #262, opened 2021-03-27
Error thrown with quick interactions


m1212e:
When I double click a draggable object I get this error:

`pointerAction.js:180 Uncaught TypeError: Cannot read property 'items' of undefined
    at HTMLElement.handleDraggedLeft (pointerAction.js:180)
    at dispatchDraggedElementLeftContainerForNone (dispatcher.js:69)
    at andNow (observer.js:81)`

isaacHagoel:
Can you reproduce it in the REPL and send a link?

m1212e:
It actually is reproduceable in the REPL in the Git Readme. I get this after some trying:
`Error:
message: "Cannot read property 'items' of undefined"
stack: TypeError: Cannot read property 'items' of undefined
at HTMLElement.handleDraggedLeft (eval at handle_message (about:srcdoc:13:8), <anonymous>:1628:67)
at dispatchDraggedElementLeftContainerForNone (eval at handle_message (about:srcdoc:13:8), <anonymous>:709:21)
at andNow (eval at handle_message (about:srcdoc:13:8), <anonymous>:1229:17)`

So what I do is, I pick an element and then keep clicking the left mouse button which starts dragging. Then I slightly move the mouse so I always start dragging a little bit. In most cases there is no error on the official REPL but sometimes it works and throws an error. I can't really identify any reason or pattern behind the error. In my particular setup it just gets triggered a lot more often.

isaacHagoel:
i tried reproducing with no success. can u provide more details?
which browser?
which OS?
which REPL on the readme?
which list on that REPL?
which element on that list (if it matters)?
are u rapidly clicking all the way through the interaction?

m1212e:
1. Brave
2. Win 10
3. https://svelte.dev/repl/e2ef044af75c4b16b424b8219fb31fd9?version=3.22.2
4. Happens on multiple ones
5. I don't think so.
6. Yes, probably

I made a short screencapture, hopefully you'll have better understanding of what I do. With every drag I do with the mouse, I do a doubleclick.
https://user-images.githubusercontent.com/14091540/112735176-b57aae00-8f4a-11eb-8580-3c42a03e53fd.mp4

isaacHagoel:
Thanks. Looks exactly like what i tried. Does it reproduce for you on
chrome or only on brave?

m1212e:
Also on a fresh installation of Chrome (no addons, nothing):
`Error:
[...]

PR #264, opened 2021-03-28
262 - added a quick exit to the drag left handler so it can't kick in if drag hasn't started

isaacHagoel:


isaacHagoel:
fixes #262

PR #269, opened 2021-04-02
Adds customStartEvent option

luizcarlos1405:
This allows specifying a custom event to start the `pointerAction` drag. I needed to start dragging only if the user holds the draggable node for some time. I used a custom `longpress` action to do so.

The custom event has to have the prop `startDragPosition` into its `details` containing `x` and `y` coordinates for the initial position. The update is quite simple, but I didn't log errors nor write tests. I'd like to know what you think first.

If this is something we wish to have in the lib (I obviously think we should), I can also implement some `customEndEvent`.

You have to pass the event and inject the event into the draggable nodes with a custom action:
```svelte
<div
  use:dndzone="{{
    items,
    customStartEvent: 'longpress',
  }}"
  on:consider
  on:finalize
>
  <span use:longpress>
    CONTENT
  </span>
</div>
```

### DIFF
--- a/src/action.js
+++ b/src/action.js
@@ -52,6 +52,7 @@ function validateOptions(options) {
         transformDraggedElement,
         autoAriaDisabled,
         centreDraggedOnCursor,
+        customStartEvent,
         ...rest
     } = options;
     /*eslint-enable*/
--- a/src/pointerAction.js
+++ b/src/pointerAction.js
@@ -310,7 +310,8 @@ export function dndzone(node, options) {
         dropTargetStyle: DEFAULT_DROP_TARGET_STYLE,
         dropTargetClasses: [],
         transformDraggedElement: () => {},
-        centreDraggedOnCursor: false
+        centreDraggedOnCursor: false,
+        customStartEvent: ''
     };
     printDebug(() => [`dndzone good to go options: ${toString(options)}, config: ${toString(config)}`, {node}]);
     let elToIdx = new Map();
@@ -367,6 +368,16 @@ export function dndzone(node, options) {
         currentMousePosition = {...dragStartMousePosition};
         originalDragTarget = e.currentTarget;
         addMaybeListeners();
+    }
+
+    function handleCustomEvent(e) {
+        originalDragTarget = e.currentTarget;
+
+        if (e.detail && e.detail.startDragPosition) {
+            dragStartMousePosition = e.detail.startDragPosition;
+            currentMousePosition = {...dragStartMousePosition};
+            handleDragStart();
+        }
     }
 
     function handleDragStart() {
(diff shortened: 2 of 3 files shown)


isaacHagoel:
Hi,
Thanks for this. Can you help me understand the use-case better?
What other scenarios do you this used for besides long-press? 
Is there a benefit to using an event rather than exposing the function that start a drag (so you could `import {startPointerDrag} from 'svelte-dnd-action'`?

luizcarlos1405:
Hmm... I don't think this is better than exposing the function. But it was the approach I found changing less code. Any other case that you could want to manually start the drag can just use the function. After reading a bit of the code Idk how to do that properly, but I'll look into it soon.

luizcarlos1405:
One way or another we need to be able to cancel the normal events. So what if we have an option `startPointerDragEvents` that takes an array of strings with the events that will start the drag and it defaults to `['mousedown', 'touchstart']`. If one wishes to cancel them they have to pass `startPointerDragEvents: [ ]` in the options. I think when we document this it will make clearer which events will, by default, start the drag. Same for `endPointerDragEvents` that would default to `['mouseup', 'touchend']`.

I don't think a `updatePointerDragEvents` would be necessary once moving with the pointer/touch is the goal of the `pointerAction`.

This change would be enough for my case but one could, in the future, implement a function to manually start the drag anywhere.

isaacHagoel:
Before we form an opinion about solution A or B it would be helpful to have a few examples of how one might want to use it in our minds.
longpress is one such example. What other usecases do you see here?

luizcarlos1405:
* Start the drag with another mouse button other than the left button
* Cancel the drag when something changes or come from the server
* Start the drag only if pressing key modifiers (shift, ctrl, alt)
* Start the drag on mouseup instead of mouedown
* Start the drag on mosedown and stick with the drag even after mouseup and just release with another mousedown

isaacHagoel:
interesting. if we want this level of customisability I think we need to store a whole set of event names for pointer and keyboard in an object and have the library use that by default.
The option can then receive an object and allow overriding some of the event names with custom event names.
Something like:
```
const DND_EVENT_NAMES = {
   POINTER: {
        MAYBE_DRAG_START = ["mousedown", "touchstart" ],
        DRAG_START: "dragstart"
        ...
   },
   KEYBOARD: {
      ...
  }
}
```
This also means the order in which triggering one event leads to the next one on the chain.
thoughts?

isaacHagoel:
@luizcarlos1405 it occurred to me (shower thought :)) that it is possible to support longpress without adding new features, using a similar trick to drag handles.
I've made a quick and dirty example for you that's based on this idea. See this [REPL](https://svelte.dev/repl/7e28249932d84d59a9d01504bf14417e?version=3.37.0)
Is this roughly what you are after?

luizcarlos1405:
Hi! I'll be available mostly on the weekend. So I like the idea of the config object, it gives the flexibility I think fits the lib. Also, this REPL does it fine! It's a little more verbose and workaroundish but it works. I'm already using my fork because the project I'm doing is fairly simple. But I think this REPL should be in the README file with the other examples, it's nice.

Do you think it's best to leave it this way? I'm still in favor of using the DND_EVENT_NAMES object to properly cover the situations I listed (and probably some more as well).

PR #271, opened 2021-04-03
270 - added option do disable morph

### LINKED ISSUE #270: [Feature] Option to disable morph
Add option to disable dragged item morph to zone dimensions. Effect is nice, but there may be cases, when you don't want morph to happen. At least our UX designer designed D&D like so. When morph is ussed, some items may be hidden under full width morphed item, also morph animation in combination with another "placeholder" shadow item animation may be disturbing. It may also be better option in combination with center on cursor => dragging smaller box.

In short: I have some palette of components (zone) to be dragged to another zone and I want these components to preserve their original dimensions.

one image === thousand words:

image

mataslib:
resolves #270


src/pointerAction.js
```diff
@@ -488,9 +490,11 @@ export function dndzone(node, options) {
             const draggableEl = node.children[idx];
             styleDraggable(draggableEl, dragDisabled);
             if (idx === shadowElIdx) {
-                morphDraggedElementToBeLike(draggedEl, draggableEl, currentMousePosition.x, currentMousePosition.y, () =>
-                    config.transformDraggedElement(draggedEl, draggedElData, idx)
-                );
+                if (!morphDisabled) {
```
isaacHagoel, line 493: please set `config.morphDisabled = morphDisabled` as some point. this will make it available elsewehre in the code if needed (by retrieving from dzToConfig).

README.md
```diff
@@ -109,6 +109,7 @@ An options-object with the following attributes:
 | `flipDurationMs` | Number | No | `0` | The same value you give the flip animation on the items (to make them animated as they "make space" for the dragged item). Set to zero or leave out if you don't want animations |
 | `type` | String | No | Internal | dnd-zones that share the same type can have elements from one dragged into another. By default, all dnd-zones have the same type |
 | `dragDisabled` | Boolean | No | `false` | Setting it to true will make it impossible to drag elements out of the dnd-zone. You can change it at any time, and the zone will adjust on the fly |
+| `morphDisabled` | Boolean | No | `false` | Dragged item is morphed to target zone dimensions by default. You can prevent this by setting this option. |
```
isaacHagoel, line 112: Maybe we can make this a bit more accurate. something like:
`By default, when dragging over a zone, the dragged element is morphed to look like it would if dropped. You can prevent it by setting this option.`
The current explanation suggests that only the dimensions change but in fact a lot of styles are copied. 
It also suggest it is morphed to the target zone but in fact it is target index within the zone and doesn't happen when it would go back to its original place (when dragged outsize of any zone).

isaacHagoel (changes requested):
thanks for this! Looks good.
i added two tiny comments.
if you also want to bump the minor version in package.json and update release-notes.md feel free to do so. Otherwise I will do it after I merge.

src/pointerAction.js
mataslib: Good point. Wilco.

README.md
mataslib: I do aggree, thanks for explanation.

isaacHagoel (approved):

Issue #300, opened 2021-06-30
styles break when used with Shadow DOM/customElements


sidvishnoi:
Example: https://codesandbox.io/s/xenodochial-thunder-cu4sj?file=/index.js (live: https://cu4sj.csb.app/)
(using JS/CSS output generated from https://svelte.dev/repl/1477eac309954f07a6c99488e7f6ada6?version=3.38.3)

This happens because the `draggedEl` is moved out from Shadow DOM to Light DOM, but the styles of child elements are left behind in the Shadow DOM. Relevant code snippet: https://github.com/isaacHagoel/svelte-dnd-action/blob/1d6fc05459d367db04544cd5bb05bfdc054bad1d/src/pointerAction.js#L393-L398

Replacing `document.body.appendChild` with `getDropZoneRoot(originDropZone).appendChild` fixes the issue, where:
```js
function getDropZoneRoot(node) {
  const rootNode = node.getRootNode(); // could be a ShadowRoot or HTMLDocument
  return rootNode.body || rootNode;
}
```

@isaacHagoel Is this something you'd appreciate a PR for?

isaacHagoel:
Would totally appreciate a PR.
Thanks

isaacHagoel:
deployed version 0.9.9, that includes the fix

PR #301, opened 2021-07-01
300 - ensure children styles don't break with customElements

sidvishnoi:
Fixes https://github.com/isaacHagoel/svelte-dnd-action/issues/300


isaacHagoel:
Looks good. I will test with the intention of merging during the weekend. Thanks

Issue #277, opened 2021-04-14
Parent node tabindex


thousandsofraccoons:
Is it possible to turn off tabindex for a node where the directive was attached? As I understand parent `<section>` doesn't have any keyboard interactions?

![image](https://user-images.githubusercontent.com/33694447/114724004-02eb6f00-9cf0-11eb-9b3e-78c9df997ac1.png)

isaacHagoel:
The parent has tab index because it has screen reader interaction: as the
user tabs in they are given info about the list.
Why would you like to remove it?

thousandsofraccoons:
I trying to make keyboard navigation. When a user presses space on the focused board wrapper nothing happens, (it has no focus styles and `outline: none`). It becomes a void step in tab navigation of the app

thousandsofraccoons:
Also, when the card is selected and click event is added, can I invoke this click event from the keyboard? Let's say I open a modal with details on click. Currently, when I press enter key - it activates keyboard dragging.

![image](https://user-images.githubusercontent.com/33694447/114804395-12050800-9d56-11eb-876a-46d234b10868.png)

isaacHagoel:
about the navigation - as i said, it is in order for the screen reader to
describe the list to the user. i think it tells the user to tab into an
item in order to drag.
i was thinking about making tabindexes configurable but i didn't end up
doing that. I guess we add this kind of functionality if needed.

about clicking - that's interesting. What is the expected behaviour?

thousandsofraccoons:
I guess I gotta try to use a screen reader myself to understand. So far as I understand, it's not required for non-interactive elements to be focusable: [https://www.w3.org/TR/wai-aria-practices/#kbd_generalnav ](https://www.w3.org/TR/wai-aria-practices/#kbd_generalnav ) Probably I must try to use a screen reader myself.

About clicking - maybe if "Space" is pressed - turn on dragging mode, if not - invoke callback and pass an event there? Would be nice to have a chance to open details on "Enter" and card to selection on "Ctrl + Enter" or show actions dropdown on some other shortcut.

isaacHagoel:
i know it is not required but it doesn't seem to make much sense otherwise. I agree it needs to be more customisable just need to figure out in what way (happy to brainstorm).
about space and click, this is a really interesting problem. there was another request - to customise the events that initiate drag start/ drop etc. Maybe this is what we need.
[...]

PR #315, opened 2021-08-30
feat: update to allow user to set custom tabindex when not dragging

wwww09z:


src/keyboardAction.js
```diff
@@ -266,6 +266,7 @@ export function dndzone(node, options) {
         items = [],
         type: newType = DEFAULT_DROP_ZONE_TYPE,
         dragDisabled = false,
+        zoneTabIndex = 0,
```
isaacHagoel, line 269: please initialise on line 159 as well

isaacHagoel (changes requested):

isaacHagoel:
resolves #277

PR #350, opened 2022-02-05
348 - cleaning up the aria divs if all the dnd zones were destroyed and initializing only when the first one is initialized

### LINKED ISSUE #348: aria attributes added to every page when using SvelteKit
Firstly, Thank you for making this great package. :)

When using svelte-dnd-action within a SvelteKit app, I'm seeing the aria attributes and alerts get added to every page:

image

This is not ideal, as Google thinks it should be the description of my website instead of the stuff I provide.

I am currently importing the svelte-dnd-action under a component used on 2 pages, however the alerts elements are being added on all pages, regardless of whether I have visited a page using the component.

Is this normal, and is there a way to not have them get added on every page? I chose to set autoAriaDisabled to true for now, but it's not a great solution.

isaacHagoel:
cleaning up the aria divs if all the dnd zones were destroyed and initializing only when the first one is initialized
fixes #348

### DIFF
--- a/src/helpers/aria.js
+++ b/src/helpers/aria.js
@@ -47,6 +47,15 @@ export function initAria() {
     }
     return {...INSTRUCTION_IDs};
 }
+
+/**
+ * Removes all the artifacts (dom elements) added by this module
+ * @return {null}
+ */
+export function destroyAria() {
+    if (isOnServer) return null;
+    [ALERT_DIV_ID, ...INSTRUCTION_IDs].forEach(id => document.getElementById(id)?.remove());
+}
 function instructionToHiddenDiv(id, txt) {
     const div = document.createElement("div");
     div.id = id;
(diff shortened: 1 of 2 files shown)


jeffalo:
I can't navigate back to a page that doesn't use the action. I just see a blank screen and this error message:

image

This occurs in production and development

isaacHagoel:
Sorry, I made a silly mistake. pls pull latest and try again.

jeffalo:
It works fine when I navigate from a page without it, to a page that uses it. But if I directly load a page that uses it, the accessibility divs don't get added, and I get this error:
image

isaacHagoel:
Alright alright i can't be lazy. I will have to carve time and actually
test it properly within a sveltekit project. Got it

isaacHagoel:
@jeffalo based on the error message you've included it looks like you were using `alertToScreenReader` which was relying on these divs being there in order to function. I made it more robust and tested locally on a svelteKit app. Please pull and try again. should be good now.

PR #375, opened 2022-05-11
Expose the dragged-over index in the DRAGGED_OVER_INDEX event

pal03377:
Currently, the DRAGGED_OVER_INDEX event doesn't actually expose the index of the item which the dragged-item was dragged over. That behavior is strange because the dragged-over index is pretty important when looking at the DRAGGED_OVER_INDEX event.

### DIFF
--- a/src/pointerAction.js
+++ b/src/pointerAction.js
@@ -207,7 +207,7 @@ function handleDraggedIsOverIndex(e) {
     const shadowElIdx = findShadowElementIdx(items);
     items.splice(shadowElIdx, 1);
     items.splice(index, 0, shadowElData);
-    dispatchConsiderEvent(e.currentTarget, items, {trigger: TRIGGERS.DRAGGED_OVER_INDEX, id: draggedElData[ITEM_ID_KEY], source: SOURCES.POINTER});
+    dispatchConsiderEvent(e.currentTarget, items, {trigger: TRIGGERS.DRAGGED_OVER_INDEX, id: draggedElData[ITEM_ID_KEY], source: SOURCES.POINTER, index});
 }
 
 // Global mouse/touch-events handlers


isaacHagoel:
The index wasn't included because i wanted the events to be consistent and adhere to "single source of truth" (as in not bloat the event with info that can be derived from info already provided in another form) - it is easy to get the index by finding the id in the items array (both are already included). Why not use that? 
With that said, I am open to adding information but we should consider (pun unintended) the other events and update the README accordingly.

pal03377:
Thanks for your thoughtful comment (and for this great library in general)! I tried finding the ID in the items, but that did not quite work as I disabled the order-switch functionality by copying the original element like on the bottom in the [Copy on drag - simple and Dragula like](https://svelte.dev/repl/924b4cc920524065a637fa910fe10193?version=3.24.1) example.

Therefore, the index of the dragged item did not change and I had no way of accessing the index of the element which my copy item was dragged over: The ID would always be constant because I'm only allowing to drag the copy.

(I'm trying to make a small educational term combination game)

https://user-images.githubusercontent.com/9006596/167951102-69db4e8a-e640-4a29-a663-ed67981b3354.mp4

In the example above, the index by finding would remain 0 at all times.

pal03377:
So essentially I needed a way to both have the index of the dragged item and the item it was dragged over, without reordering being enabled. Maybe there is another elegant way to get both of those without internal changes?

isaacHagoel:
Are you searching the index of the id in the items provided in the event
(should have the correct index) or the items in your component (should have
the wrong index in your case)?

pal03377:
Oh, you are absolutely right! Sorry, I didn't consider that. In that case, my PR is completely redundant :-)

Thank you so much for your help!

PR #428, opened 2023-01-29
Fix scroller logic

null-dev:
The original logic checks if the element is currently scrolling, if so, it will not begin another scroll. However, because it is missing a return, not only will it not begin another scroll, it will also proceed to the end of the function where it will call `resetScrolling()`. This terminates the ongoing scroll. So even though the element needs scrolling, no scrolling will be done at all.

This change adds the missing return.


isaacHagoel:
thanks for that. how do i test the before and after?

Issue #305, opened 2021-07-21
Constrain drag axis


inwardmovement:
Is it possible to constrain drag axis, for example to allow drag only on the vertical axis for vertical lists, whatever the cursor position?
I think it would give a better UX.

![image.png](https://user-images.githubusercontent.com/9438102/126470099-33a38193-834a-4512-8199-6257bdadbfd3.png)

isaacHagoel:
It is not currently supported but shouldn't be very difficult to add.
Is it for a single list usecase?

inwardmovement:
Yes for a single list use case as far as I'm concerned, but it should work in all cases I think (be a "general" drag and drop/UX behaviour).

Any idea how we could achieve that?

isaacHagoel:
It will require enhancing the library. There is a piece of code (in
pointerAction.js if i recall) that updates the transform of the dragged
element based on the mouse movement. It will need to only update the x or y
coords when this option (new option) is enabled

inwardmovement:
Are you willing to add this feature?
Do you want me to try a PR?

isaacHagoel:
If you want to give it a shot it would be great. If not, i will try to make
time for it

inwardmovement:
I think it's on lines 217-219 of `pointerAction.js`, but I can't manage to make it work. I tried with the first parameter at `0` but it doesn't change anything.

isaacHagoel:
You've found the right place. Changing line 217 to 
```javascript
draggedEl.style.transform = `translate3d(0, ${currentMousePosition.y - dragStartMousePosition.y}px, 0)`;
```
results in locking the initial X coords (see capture). Are you sure you used the modified version (did u use npm link to a local copy)?
https://user-images.githubusercontent.com/20507787/126577993-5ca408ff-2b25-4712-b506-5266a7f7298f.mov

inwardmovement:
Nice! No I didn't use npm link... I thought rebuilding with another output dir was fine.

isaacHagoel:
cool. please let me know.

isaacHagoel:
closing for now as "won't do unless more people need this". still open for a MR.

ChugunovRoman:
@isaacHagoel @inwardmovement I created [PR](https://github.com/isaacHagoel/svelte-dnd-action/pull/429) by this issue

PR #429, opened 2023-02-01
Add options for constrain by each axis

ChugunovRoman:
Related #305

### DIFF
--- a/src/pointerAction.js
+++ b/src/pointerAction.js
@@ -214,7 +214,11 @@ function handleDraggedIsOverIndex(e) {
 function handleMouseMove(e) {
     e.preventDefault();
     const c = e.touches ? e.touches[0] : e;
-    currentMousePosition = {x: c.clientX, y: c.clientY};
+    const {constrainAxisX, constrainAxisY} = dzToConfig.get(e.currentTarget) || dzToConfig.get(originDropZone);
+    currentMousePosition = {
+        x: constrainAxisX ? dragStartMousePosition.x : c.clientX,
+        y: constrainAxisY ? dragStartMousePosition.y : c.clientY
+    };
     draggedEl.style.transform = `translate3d(${currentMousePosition.x - dragStartMousePosition.x}px, ${
         currentMousePosition.y - dragStartMousePosition.y
     }px, 0)`;
@@ -322,6 +326,8 @@ export function dndzone(node, options) {
         items: undefined,
         type: undefined,
         flipDurationMs: 0,
+        constrainAxisX: false,
+        constrainAxisY: false,
         dragDisabled: false,
         morphDisabled: false,
         dropFromOthersDisabled: false,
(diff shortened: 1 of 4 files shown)


src/pointerAction.js
```diff
@@ -214,7 +214,11 @@ function handleDraggedIsOverIndex(e) {
 function handleMouseMove(e) {
     e.preventDefault();
     const c = e.touches ? e.touches[0] : e;
-    currentMousePosition = {x: c.clientX, y: c.clientY};
+    const {constrainAxisX, constrainAxisY} = dzToConfig.get(e.currentTarget) || dzToConfig.get(originDropZone);
```
isaacHagoel, line 217: why does it read it from currentTarget? shouldn't it be determined by the origin only?

isaacHagoel (commented):
So what's the usecase here? should the element be allowed to leave the origin dropzone?
How is keyboard dragging affected?
How is highlighting potential drop targets affected?

ChugunovRoman:
@isaacHagoel 

> So what's the usecase here?
> 
should the element be allowed to leave the origin dropzone?

> How is keyboard dragging affected?
This has no affect on keyboard dragging. The behavior of keyboard dragging does not changed.

> How is highlighting potential drop targets affected?

ChugunovRoman:
@isaacHagoel 

> So what's the usecase here?
For example, I use this changes for tab swap by mouse and tabs should be moving only by X coordinate.
Also this feature can be applied in a board for constrain swap columns by Y.
And etc.

> should the element be allowed to leave the origin dropzone?
What does this question related to the functionality of constraiting the movement of an element along the axes by dragging with the mouse?
I think that the functionality to allow the element to leave the dropzone is a separate feature that can be implemented if necessary.

> How is keyboard dragging affected?
This has no affect on keyboard dragging. The behavior of keyboard dragging does not changed.

> How is highlighting potential drop targets affected?
On highlighting potential drop targets this does not affect too.

isaacHagoel:
I'll try to explain my concern better. Let's say I have 4 dropzones of the same type organized this way:
             U
L.                       R
             D
If you limit L to the X axis that means that elements can only be moved to R if you are using the mouse/touch but anywhere if you use the keyboard.
Also, when lifting an element from L all dropzones will be highlighted indicating that they are drop targets when in fact they are not. Same issue with the aria messages.
If L happens to be a vertical list, it also means that the items cannot be re-sorted and if R is vertical it means that any element taken from L and moved to R has only one index it can be dropped into.
This doesn't feel like good DX or UX to me. I am open to such feature, generally speaking but it needs to be thought out.

ChugunovRoman:
I think I understood. Are you talking about the behavior of the element that was dragged out of the dragzone? I think we also able to change this behavior via new property in the options.

segovo:
This feature works great for me, I'd just like to add that changing createDraggedElementFrom() should be considered so that if centreDraggedOnCursor is true it doesn't move the element in one of the constrained axises. In other words, centreDraggedOnCursor should only center an item vertically in a vertical only movement constrained list.

isaacHagoel:
good point. There are many considerations and every new feature has to work
in harmony with what came before. I am glad it's working great for you but
I can't merge it in this way.

PR #430, opened 2023-02-01
Add ability set custom cursor styles for different mouse event

ChugunovRoman:
`!important` in client code bad works for draggable elements. This changes allow set custom cursor styles from all available cases

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,16 +104,20 @@ npm install --save-dev svelte-dnd-action
 
 An options-object with the following attributes:
 | Name | Type | Required? | Default Value | Description |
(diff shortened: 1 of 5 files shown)


isaacHagoel (commented):
The footprint of this feature is too large for what it does and given that there is already a way to achieve the same thing (which is admittedly not the cleanest)

ChugunovRoman:
> The footprint of this feature is too large for what it does and given that there is already a way to achieve the same thing (which is admittedly not the cleanest)

Do you mean to toggle element classes from code instead of change defined styles? And then user able to overwrite styles for setted classes? Or can you offer another solutions?

PR #453, opened 2023-06-23
Fix wrong id type in DndEventInfo

shueja-personal:
When logging `e.detail.info` it shows that `id` is a number, just like in the items list. However, the typings show `DndEventInfo.id` to be a string, which is incorrect.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -66,7 +66,7 @@ export enum SOURCES {
 
 export interface DndEventInfo {
     trigger: TRIGGERS; // the type of dnd event that took place
-    id: string;
+    id: number;
     source: SOURCES; // the type of interaction that the user used to perform the dnd operation
 }
 


isaacHagoel:
the id would be whatever was passed in to the lib, no? as in if u pass in a string it will be a string if u pass in a number it will be a number

shueja-personal:
The examples all use numbers, so I figured that `id` was required to be a number. Looking at the typings it's not even enforced that `Item` has an id property. Either way, the typings expect `id` in DndEventInfo to be a string, with no regard to what is passed into the library. To my knowledge, there's not a super good way to make DndEventInfo expect an `id` type that matches the item type, but that would be the ideal.

isaacHagoel:
I agree that it would be ideal if possible, and i think that in most real life cases it would be a string (but enforcing it might break existing users)

Issue #403, opened 2022-10-05
Transform element dragged from another dnd zone


madeleineostoja:
I'm trying to find a solution to transforming a dragged element once it enters another dnd zone of the same `type`, is that something the current API of `svelte-dnd-action` can address? Because I can't figure it out. `transformDraggedElement` seems to only be called from the originating dnd zone and can't differentiate when the element enters another dnd zone (can it?).

To further explain the use-case for this, I have: 
1. A grid of photos which is a dnd zone (to rearrange items)
2. A list of galleries which is a dnd zone (to rearrange items)
3. Individual galleries in the list of galleries are **also** a dnd zone with the same `type` as the grid of photos, so I can drop photos into the gallery list to move them to that gallery

What I want:
- To transform a photo when it is dragged over a gallery item, to show it can be moved/dropped there.

Video demonstrating what I mean:

https://user-images.githubusercontent.com/6374876/193975018-159e581e-968b-4b15-9af6-00585a0623e2.mov

isaacHagoel:
In theory it should work. Can you pls create a REPL that reproduces the
issue?

madeleineostoja:
Sure thing, started working on a repl but I'm getting an error I'm not seeing in my dev environment, I'll debug it when I get a mo unless you know what I'm doing wrong offhand

https://svelte.dev/repl/2b0ac0549af5432a9cfc2bbac3dc3b53?version=3.50.1

madeleineostoja:
Okay I think I've figured it out, this is another face of #374. Since the item I'm dragging into doesn't have/display children, it doesn't morph when dropping (and I've disabled morph because of that), so `transformDraggedElement` is never called because it's only called if morph isn't disabled ([here](https://github.com/isaacHagoel/svelte-dnd-action/blob/2cb159806449d3007c9fe7fb84538c3d0bfc4127/src/helpers/styler.js#L88) and [here](https://github.com/isaacHagoel/svelte-dnd-action/blob/2cb159806449d3007c9fe7fb84538c3d0bfc4127/src/pointerAction.js#L517))

isaacHagoel:
The REPL doesn't give me an error message, how do you make it error out?

madeleineostoja:
Drag an image into a gallery list item

![Screen Shot 2022-10-12 at 10 33 46 AM](https://user-images.githubusercontent.com/6374876/195202650-4f87e1a9-530a-4a52-9555-49884947cd89.png)

[...]

PR #469, opened 2023-08-10
Feat: "drop only" dndzones

eduardomartines:
# Motivation
I needed a zone that was only dropable so I created a dropOnly attribute in dndzone;

I think its also somehow related to things like that https://github.com/isaacHagoel/svelte-dnd-action/issues/403

Any thoughts on that? Like a better way to accomplish this?

Thank you for your time!

### DIFF
--- a/src/action.js
+++ b/src/action.js
@@ -19,6 +19,7 @@ import {toString} from "./helpers/util";
  * @property {number} [zoneTabIndex] - set the tabindex of the list container when not dragging
  * @property {object} [dropTargetStyle]
  * @property {string[]} [dropTargetClasses]
+ * @property {boolean} [dropOnly]
  * @property {function} [transformDraggedElement]
  * @param {HTMLElement} node - the element to enhance
  * @param {Options} options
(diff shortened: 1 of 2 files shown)


isaacHagoel:
what is the difference between this and the existing `dragDisabled` prop?

eduardomartines:
> what is the difference between this and the existing `dragDisabled` prop?

with `dragDisabled` the `items` is required, `dropOnly` assumes the zone is just a simple "bucket".

isaacHagoel:
But items can be empty (an empty list)

eduardomartines:
`items` with an empty array cause errors to start popping out:

1. "Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'id')"
2. "Uncaught TypeError: Cannot read properties of undefined (reading 'isDndShadowItem')"

isaacHagoel:
that would be a bug. I just tried it now with one of the examples and it
worked. can u share an example in which an empty list causes issues?

eduardomartines:
Sure, I gonna isolate this in a REPL over the weekend

Issue #479, opened 2023-08-29
Registering a new drop zone while dragging an element


m4nnke:
Hi, 
first of all great work with your library.

I have a use case, where I display the nodes of a simplified abstract syntax tree of a latex document using dropZones as inner Nodes of the graph. It's a LOT of nesting. Inner Nodes can be collapsed so parts of the tree aren't visible. Everything works quite smoothly :).

Now to the point:
While dragging an element, a new dropzone is being mounted (its part of a collapsed inner node that opens after hovering over it). The problem is, that it's not possible to drop the dragged element in the newly created drop zone. (Probably because it is not yet registered in the typeToDropZones Map or something). Is there a way to register a new drop zone while dragging an element?

Thanks in advance

isaacHagoel:
Hi,
Do you mind making a simple REPL that illustrates the issue?
Based on your description I would expect it to work (so maybe there's a bug)

m4nnke:
Thanks for the fast reply. Maybe my description was not detailed enough. I hope this illustrates my problem: 
https://svelte.dev/repl/9af9c4f11c2f4d888438b8b29d9e6c52?version=4.2.0

isaacHagoel:
(cont) the key is to call `observe` on the new zone and add the event listeners. the event listener on the window shouldn't be added again.

isaacHagoel:
@m4nnke i merged that PR but now that i am thinking about it I reckon that the opposite case is probably also an issue...
as in if the user removes a zone mid drag and it unregisters, the lib won't stop observing it and I wonder if that would cause any errors.

m4nnke:
Maybe this is linked with the potential bug I mentioned in #480 first message all the way down (occurs when removing the source zone on drag). You can try it out in the test repo I created. I'll think about that tomorrow. I'm from Europe, it's quiet late here ;)

m4nnke:
I tried my old implementation and it seems, the issue is not occuring there. I stumbled upon this issue too earlier today.

m4nnke:
I think the problem is the fact that rewatching the dragged element leads to reobservation, of the lists, wich leads to a reinitialisation of the variables inside the `observe` function. The next time the dragged element enters a dropzone things go wrong (I'm not sure what exactly). In my first impementation the observe loop keeps running the whole time so the variables stay in tact.
[...]

PR #480, opened 2023-08-30
Change observer zone list to be updatable and update it on new zone while dragging

m4nnke:
Hey, this is linked to Issue #479 
as far as I can see I have implemented a solution. My solution udates the observe method everytime a dropzone is registered, the zone has the correct type and `isWorkingOnPreviousDrag` is true. One thing a am confused about: Why is the current dropzone registered on every considering of a drag in the zone (that is the case without my changes too)? That leads to lots of updates on the observer zone list and event listeners on the currently active zone will be registered multiple times.

Furthermore I don't know if there's a problem with pushing the [registration](https://github.com/m4nnke/svelte-dnd-action/blob/b85f00dcb45f018c0e00122f7a06667a9b3feee8/src/pointerAction.js#L525C15-L525C15) method down a few lines in `configure`, is that a problem? I need the finished config (specificly the `dropAnimationDurationMs`) to update the `currentObservationIntervalMs`.
 
And I think `dropZonesFromDeepToShallow` in [observer ](https://github.com/m4nnke/svelte-dnd-action/blob/b85f00dcb45f018c0e00122f7a06667a9b3feee8/src/helpers/observer.js#L18)doesn't necessarily need to be global in this implementation. I think it could just be declared as before without being `const`.

I was confused why there were multiple registrations from keyboardAction and pointerAction, to be more meaningfull, I changed the debug output to show, where the registration came from.

I hope that helps.

Might be a bug I encounterd: In the test REPL I sent in #479: If you drag an element and close the drop zone the dragged element is from by hovering over title, then letting go of the element, there will be an error the next time you open the drop zone.

### DIFF
--- a/src/helpers/observer.js
+++ b/src/helpers/observer.js
@@ -15,6 +15,9 @@ const INTERVAL_MS = 200;
 const TOLERANCE_PX = 10;
 const {scrollIfNeeded, resetScrolling} = makeScroller();
 let next;
+let dropZonesFromDeepToShallow;
+let needsUpdate = false;
+let newDropZones;
 
 /**
  * Tracks the dragged elements and performs the side effects when it is dragged over a drop zone (basically dispatching custom-events scrolling)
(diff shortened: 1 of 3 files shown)


isaacHagoel:
Ah, in my mind this should have been simpler to implement (I haven't tried though). Sorry if I didn't explain it well enough. The new zone does register as is but it is not being watched because that function is only called on drag start on the zones that existed then. Originally I said all you need to do is call (a slightly more flexible version of) watch on the new zone, but now that I see this I realise it can be even simpler, all you need to do if drag is ongoing while registering is unwatch everything and then rewatch again and I think it will just work (`unWatchDraggedElement()` and then `watchDraggedElement()`).

m4nnke:
Ok, my bad, that is a lot simpler and works the same way. I tried that, but I did it wrong. The key is to only rewatch, when the new zone is not yet in the zones map. Otherwise it runs into an infinite loop. 
But the registration method has to fire after the config is set for the zone, as mentioned before.

I'll create a new Pull request.

isaacHagoel:
Thanks. Please also update release-notes.md and bump the version in package.json.

isaacHagoel:
this is done in a cleaner way in https://github.com/isaacHagoel/svelte-dnd-action/pull/481. closing this one.

PR #481, opened 2023-08-30
Rewatch dragged element on zone registration while dragging

m4nnke:
Hey, this is linked to Issue #479 (second try)

This solution seems to be working fine.

When rewatching the dragged element the config of the new zone has to be read, so the `registerDropZone` method has to fire after the config added to the config map.


isaacHagoel:
Looks good. I will test it just to be sure later today or tomorrow. 
If you have time please update release-notes.md and bump the version in package.json. If not, I will do it after I merge.

isaacHagoel:
fixes #479

Issue #455, opened 2023-06-23
Usage within dialog element


canastro:
The new [Dialog element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) moves things to a top-layer, I'm having troubles rendering the dragged-element when using svelte-dnd-action within a dialog.

Is this a known issue? Is there a workaround? Am I missing something?

isaacHagoel:
Not a known issue. Can you please make a REPL that demonstrates the issue?

canastro:
@isaacHagoel here it is: https://svelte.dev/repl/a7c16138992a4c1db6b0e578952145bf?version=4.0.0

isaacHagoel:
Thanks. I looked into it. 
It turns out that the dialog element is added to a special layer called "the top layer" that renders above anything else (including the dragged element which is a child of the document and cannot be a child of the list because svelte controls the rendering of the list).
More info [here](https://blog.bitsrc.io/top-layer-vs-z-index-for-frontend-dialogs-fcb12217a834) and [here](https://stackoverflow.com/questions/41815069/is-it-possible-to-position-a-div-on-top-of-a-dialog-tag-that-is-not-its-parent).
Not sure what I can do about it. Luckily it is easy enough to implement a dialog element that won't have this issue when there is a need to include a dnd zone inside. Marking this as a bug with a known workaround.

canastro:
Unfortunately, in my opinion, thats not a good workaround. The introduction of the Dialog element serves a purpose and is an advancement in webdev/accessibilit. Having some dialogs using <dialog> and others a custom implementation just to support dnd doesnt sound like a future-proof approach.

Ideally we would be able to query if we're within a top-layer and append all DND elements to the root of that layer instead of appending to the body of the document.  Right now only the Dialog uses top-layer, but soon the PopoverAPI will also use it and this issue will come up more frequently (I think).

I can't find any good way to query the root of a top-layer, so for now, maybe we could, optionally, pass a reference to the element that we want to use as root?

```svelte
<section 
  use:dndzone={{ root: dialogElement, items, flipDurationMs }} 
  on:consider={...} 
  on:finalize={...}
>
...
</section>
```

isaacHagoel:
I am open to a pull request. One thing to think about is whether there is a
need to check that the provided root element is "legal" (e.g imagine you
get one of the items in the list or whatever)

isaacHagoel:
fixed by the merged PR

PR #499, opened 2023-10-01
Support Dnd inside html dialogs

thijsmie:
Html dialogs are rendered as a #top-layer and if you append the dnd shadow element to the body it will be rendered under the dialog. I'm not sure this is the best way to support this but at least it works on my end. 😇


src/pointerAction.js
```diff
@@ -383,8 +383,8 @@ export function dndzone(node, options) {
         const currentIdx = elToIdx.get(originalDragTarget);
         originIndex = currentIdx;
         originDropZone = originalDragTarget.parentElement;
-        /** @type {ShadowRoot | HTMLDocument} */
-        const rootNode = originDropZone.getRootNode();
+        /** @type {ShadowRoot | HTMLDocument | Element } */
+        const rootNode = originDropZone.closest("dialog") || originDropZone.getRootNode();
```
isaacHagoel, line 387: i wonder if it is possible to have a shadow root inside of a dialog. if it still works if you reverse the order of the || condition that might be safer

isaacHagoel (changes requested):

src/pointerAction.js
thijsmie: I could check and change the implementation, or I could follow the strategy and the limitations you mentioned here instead: https://github.com/isaacHagoel/svelte-dnd-action/issues/455, do you have a preference?

isaacHagoel: I think your idea is simpler and less open to abuse then allowing the user to specify the root element (i hope I understood your question correctly).
I just asked the question to make sure there are no weird edge cases we are not thinking of.
If u can test with shadow root inside of a dialog it would be great (no need to reverse the || just check what happens)

thijsmie: I will check after regular work and get back to you 👍

isaacHagoel: Hi any updates on this one?

thijsmie: Hi Isaac! Sorry, this got forgotten in the daily churn of stuff. Thanks for pinging me. I've confirmed that a `.closest()` call does not propagate through a shadow DOM root (as it is treated as a document root) in latest chrome and firefox. Does that resolve your concerns enough?

isaacHagoel: yes. can you please pull from master, bump the version in package.json and update release-notes.md.
once you do i can merge

thijsmie: Here you go!

isaacHagoel (approved):

Issue #311, opened 2021-08-03
[Feature Request] Support autoscroll on scrollable parent (container)


mataslib:
The library supports autoscroll out-of-the-box for an element having use:dndzone (edges) and window (edges and mouse wheel).

However zones may often be nested in scrollable parent/container as they are in my case. It would be nice to support autoscroll for scrollable parent (as react-beautiful-dnd does by default). Or support at least for passed element/selector via options.

isaacHagoel:
what are cases in which you can't make the dndzone the scrollable element?

mataslib:
Element builder with lots of nested dndzones (section > column > element) inside single scrollable 100vh canvas which can not be body for some reason.

![image](https://user-images.githubusercontent.com/5154014/128133407-91002c99-3a66-49aa-8719-11338cdbc030.png)

gyurielf:
I ran into the same issue..
Yep! It would be nice!

TheAsianNextDoor:
I noticed in the reproduction that if you scrolled all the way down and rearranged `fixed parent 3`, there was a jump. I believe it's due to the need to refocus the top of the container element. I might be wrong tho, what else could be causing that?

isaacHagoel:
Closing as superseded by fixes and APIs now present in v0.9.77. If this still reproduces on v0.9.77, please open a focused issue with current versions and a minimal reproduction.

PR #506, opened 2023-10-26
Scrolling issue fixes

fnimick:


isaacHagoel:
Just to let u know that I saw this and will review no later than next week

isaacHagoel:
@fnimick thanks for making it and I like the overall direction.
The implementation feels a bit pervasive and I am concerned about maintaining it in the future.
Can it be more encapsulated?
For example we can make all the logic contained inside the scroller (maybe making the current mouse position globally available/ tracked globally for everyone would help) and change the rest of the code base more minimally?

fnimick:
I don't know how you would want me to do that. Can we merge the change for scrolling parent containers (which is causing a prod issue that I had to work around) and ignore the cursor position change for now?

The fundamental problem is that the library passes around the dragged element object, and then introspects its current state center position for all calculations, including scrolling. We would have to track a second state element if we don't want to simply store the mouse offset of drag start from the element center.

isaacHagoel:
yes. if it is urgent to you I am okay with the scrolling the correct parent part. 
you can make it it's own PR. 
If you do please don't forget to update the release notes, readme if needed & bump the version in package.json
make sure yarn test passes and make a REPL we could add to the examples in the README showing nested containers scrolling

About the other part - I understand the issue. i will give it some more thought. maybe what we actually need is the option to choose whether to use the pointer rather than the center for all calculations.

fnimick:
@isaacHagoel I'll make a separate scrolling PR and send that tomorrow.

re: "maybe what we actually need is the option to choose whether to use the pointer rather than the center for all calculations" - I tried that actually and it makes the behavior of where elements end up in a dragged list feel really weird. You would expect list items to end up before or after based on the dragged element itself, not based on the cursor.

I can't think of any instance where you would want scrolling to _not_ be based on the cursor - with it based on the element center if you grab a moderately tall element near its top edge, your cursor will leave the viewport and drop the element before the vertical centerline gets close enough to the top to trigger the scroll behavior (this is the problem I ran into that prompted my fix)

isaacHagoel:
Maybe we should scroll when the edge of the element goes out of the edge of
the viewport...

fnimick:
That potentially opens up a new failure case to contend with: what happens when you grab a tall element that already has partially scrolled out of view?

isaacHagoel:
maybe in that case u just scroll right away (because when you check u find
that the edge is already out of the view port). it is done by sampling the
state of the dom in every cycle so I think it would work fine.
not sure if it will be better or worse UX comparing to pointer location.

fnimick:
@isaacHagoel  this PR is now only the scroll parent container fix. (We should probably add a cypress test to verify that behavior at some point, but I do not have the time.)

src/helpers/observer.js
```diff
@@ -36,7 +52,12 @@ export function observe(draggedEl, dropZones, intervalMs = INTERVAL_MS) {
      */
     function andNow() {
         const currentCenterOfDragged = findCenterOfElement(draggedEl);
-        const scrolled = scrollIfNeeded(currentCenterOfDragged, lastDropZoneFound);
+        const scrollableContainers = findScrollableContainers(lastDropZoneFound);
+        let scrolled = false;
+        for (const container of scrollableContainers) {
+            // scroll only the first container that needs to be scrolled
```
isaacHagoel, line 58: if only the first element needs to be scrolled why not `break` after you found one that returned true instead of letting the loop continue to run? I understand your code does the same implicitly but why not make it explicit?
(`if (scrolled) break` or something like that)

src/helpers/observer.js
```diff
@@ -36,7 +52,12 @@ export function observe(draggedEl, dropZones, intervalMs = INTERVAL_MS) {
      */
     function andNow() {
         const currentCenterOfDragged = findCenterOfElement(draggedEl);
-        const scrolled = scrollIfNeeded(currentCenterOfDragged, lastDropZoneFound);
+        const scrollableContainers = findScrollableContainers(lastDropZoneFound);
```
isaacHagoel, line 55: do we need to find the scrollable container in every cycle? it would be much cheaper (cpu) before we call "andNow". Doable?

src/helpers/observer.js
```diff
@@ -16,6 +16,22 @@ const TOLERANCE_PX = 10;
...
 
+function findScrollableContainers(element) {
+    if (!element) {
+        return [];
+    }
+    const scrollableContainers = [];
+    let parent = element;
+    while (parent) {
+        const {overflow} = window.getComputedStyle(parent);
+        if (overflow.split(" ").some(o => o === "auto" || o === "scroll")) {
```
isaacHagoel, line 27: can "overflow" ever be undefined (on any browser etc)?

isaacHagoel (commented):
Thanks for this.
1. A question about how this functions...
If there are multiple scrollable containers nested within each other it seems like this would scroll the first but when it is impossible to keep scrolling it it would start scrolling the next one. is that correct? what is the benefit of scrolling anything but the closest parent?
2. please don't forget to update the release notes, readme if needed & bump the version in package.json. Make sure yarn test passes and make a REPL we could add to the examples in the README showing nested containers scrolling

fnimick:
Re: why I scroll multiple parents - say there are two drop zones that are scrollable, inside a parent that is also scrollable. If we cut off scrolling at the first one, then we can't scroll the parent container to the second drop zone in order to move the item to it.

I'll address the other comments and do version bumps, etc today. (though how should I make a REPL before it's published to npm?)

isaacHagoel:
Thanks. Makes sense.
Regarding the REPL, I don't expect it to be a REPL per se until we publish
but I need the code that would go to that REPL (basically the code you are
using to test your change)

isaacHagoel:
Hi, any updates on this (please pull from master when proceeding)?

fnimick:
I ended up getting pulled into other project, but I'll look in the next few days.

Re: scrolling based on the edge of the element - I still think scrolling based on mouse position relative to the edges of the scrollable container makes the most intuitive UX sense. Is there an easy way to pass that down to the element observers in a way that it updates live with mouse movements? I used the offset mechanism since the element reference can be used to get its current position, and then we can easily calculate the current mouse position from the element's current position + the offset provided at drag start.

Actually, I wonder what happens if an element attempts to resize itself while being dragged... I'll have to look at that case as well.

isaacHagoel:
Following up. I don't mind doing it in a new PR (the scrolling parents bit for now) but then you won't appear as a contributor and I think that would be a shame and unfair

src/helpers/observer.js
fnimick: not according to the spec - it's always a defined string value and defaults to `auto`

fnimick:
@isaacHagoel finally getting back to look at this.

I found a potential problem. Suppose we have two separate drop zones inside a large scrollable parent container. As soon as the element leaves the first dropzone, the scrolling calculation stops because the "last dropzone found" is now undefined. This is a library approach and UX question: should we be able to scroll arbitrary page containers by dragging over them, because perhaps scrolling that container may lead to a dropzone becoming visible that is a valid target for the dragged element? Or should we do a calculation based on the all possible scrollable dropzone parents and then limit based on whether the element is currently inside that container before doing the scroll check?

isaacHagoel:
It makes sense to allow scrolling any scroll container that is a parent of
a drop zone. This is why there is also a "window scroller".
Another note: I noticed that elsewhere in the code the test as to whether
scrolling is possible uses a comparison of the client height/width to the
scroll height width. For consistency maybe use that rather than looking at
the computed styles.

fnimick:
Makes sense, in that case this code doesn't solve that problem. It will only scroll the parent of the drop zone as long as the dragged element is within it. As soon as it leaves, we set the "last drop zone" to undefined and no longer calculate parents.

Is the solution to calculate all possible parents of all drop zones and scroll those when the element enters them and is near the edge?

isaacHagoel:
yeah sounds like

dusty-phillips:
@fnimick is this ready as of your most recent push, or is more work needed to solve the parent calculating issue?

isaacHagoel:
@dusty-phillips this has been stale for a while. what do you need?

isaacHagoel:
if you can, pls create a new issue for this and include a REPL reproducing
it and I'll have a look

dusty-phillips:
I think I was going off of a comment in this example:

https://svelte.dev/repl/288f827275db4054b23c437a572234f6?version=3.38.2

In `Column.svelte` there is a line that says:

```
        /* Notice that the scroll container needs to be the dndzone if you want dragging near the edge to trigger scrolling */
```

This comment may be out of date because in the repro I tried to generate, scrolling on drag (both groups and items) is working fine:

https://svelte.dev/repl/4675f90eca4b46f5a914b0d0f1f5b90f?version=4.2.12

It is *not* working in my production app, however. I will comb through it and try to identify any differences between it and the working version I linked here. Outcome will either be a fix in my code or a new issue and repro here. I'll let you know either way.

fnimick:
Yes, you can't scroll the parent container of a dndzone other than the window, without this fix.

It was ready to merge at one point but has fallen out of date since we disagreed on the approach for calculating pointer position for triggering scrolling.

dusty-phillips:
I didn't make a new issue as #311 already covers it, but I did comment there with a summary. In fact the repro I linked above does not scroll properly on chrome.

isaacHagoel:
alright. I picked it up and will address the last points that were raised
in the discussion. hopefully i will have something to share (a new PR) this
weekend.

isaacHagoel:
I created #549, it scrolls parent containers as well as the viewport. It is still not 100% ready but it works on my machine ;)
@dusty-phillips @fnimick please have a look/ take if for a spin if you can.

isaacHagoel:
closing this one as #549 was just merged

fnimick:
Confirmed that the new change fixes my issue, and your way is much more elegant. thank you!

isaacHagoel:
@fnimick thanks for confirming. It was possible due to the discussion I had with you and using your ideas as reference.

PR #508, opened 2023-10-29
update for svelte 4.2

ryoppippi:
From 4.2, we can declare 'svelte/elements' for custom attributes.
I use this and this is so good!

https://svelte.jp/docs/typescript#enhancing-built-in-dom-types

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,13 +269,12 @@ declare namespace svelte.JSX {
 #### Svelte 4:
 
 ```typescript
-declare type Item = import("svelte-dnd-action").Item;
-declare type DndEvent<ItemType = Item> = import("svelte-dnd-action").DndEvent<ItemType>;
-declare namespace svelteHTML {
-    interface HTMLAttributes<T> {
-        "on:consider"?: (event: CustomEvent<DndEvent<ItemType>> & {target: EventTarget & T}) => void;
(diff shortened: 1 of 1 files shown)


fnimick:
@ryoppippi does this mean that modifying the global types for `HTMLAttributes` is no longer necessary? If so, there should be an indication in the typescript support section of the README that the modification is not necessary if svelte 4.2+ is used.

isaacHagoel:
does this work for versions >=4.0 && < 4.2, based on the title I assume no but the README just says Svelte 4 so it might not work for some. Also since most people simply copy paste it and move on I wonder if it is worth keeping it as is just to save the mental overhead of "wait do I need to copy this or that?" and "wait, I am going to upgrade to 4.2 later, will I need to change this?"

fnimick:
So the reason would be that if the signature of the events from this library changes, or a new event is added in the future, users won't have to manually update their vendored type definitions to keep them in sync.

ryoppippi:
Ok, we need to add like `require 4.2`

ryoppippi:
If the HTMLAttribute type could be defined in the library and automatically loaded at import, it would eliminate the need for the user to copy and paste the type.

ryoppippi:
If so, I'll create another PR

PR #510, opened 2023-11-02
Update README.md

VanillAquila:
Screenshot from 2023-11-03 00 10 42
Everywhere in code, including dockblocks and Readme.md Default value column that is up to date. Except this lesser part


isaacHagoel:
Yeah this readme section is not clear I agree but zero has a special meaning nontheless.
See https://github.com/isaacHagoel/svelte-dnd-action/pull/493/files
If it is unset the lib "reaction time" will be 100ms (and no animation on drop, unlike when u explicitly set it to 100).
If you set it explicitly to zero it becomes 20ms.
higher values (than a 100) are possible and should be used when you are using slower flip animations.

VanillAquila:
gocha, sorry for bothering, and thanks a lot for explanation :+1:

Issue #526, opened 2023-12-16
ReferenceError: self is not defined after migrating to sveltekit2


noobmaster19:
Here's a shortened form of the error:

```
4:16:44 PM [vite] Error when evaluating SSR module 
...
ReferenceError: self is not defined
    at .../.pnpm/svelte-dnd-action@0.9.34_svelte@4.2.8/node_modules/svelte-dnd-action/dist/index.js:4:23
    at .../.pnpm/svelte-dnd-action@0.9.34_svelte@4.2.8/node_modules/svelte-dnd-action/dist/index.js:5:2
    at instantiateModule 
```

isaacHagoel:
How did you encounter this? By upgrading sveltekit? Was it working with the
previous version of sveltekit (or the lib)? Which line is it complaining
about (can you click into it)?
In general SSR should ignore the lib altogether

Rar9:
+1 getting the same  [vite] Error when evaluating SSR module 
ReferenceError: self is not defined

Rar9:
down grading from 0.9.34 to 0.9.33 helps but gives this warning

[vite-plugin-svelte] WARNING: The following packages have a svelte field in their package.json but no exports condition for svelte.

svelte-dnd-action@0.9.33

Please see https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#missing-exports-condition for details.

isaacHagoel:
@Rar9 yeah that's the warning 0.9.34 was supposed to fix. checking what this new error is about

isaacHagoel:
I don't have time to get to the bottom of this now (hopefully tomorrow). The lib itself doesn't use "self" (it does use "window" quite a lot) so this error has to do with the code that rollup adds (the irony :)) as can be seen [here](https://www.unpkg.com/browse/svelte-dnd-action@0.9.34/dist/index.js). Version 0.9.33 has that too, see [here](https://www.unpkg.com/browse/svelte-dnd-action@0.9.33/dist/index.js) so I am not sure why it doesn't trigger the same error. This wasn't a problem with Sveltekit 1 so very strange indeed. 
This means that changes to the library code will not solve the issue. It has to be something in the rollup config or the exports config or god knows what. 
I followed the advice on the link with 0.9.34 thinking it is a quick fix but looks like I was wrong. 
Use 0.9.33 for now. It has all the latest feature. 

I will try to make some time for this tomorrow. If anyone knows the solution please do let me know :)

[...]

PR #528, opened 2023-12-18
fix: add import and require exports

iamsbro:
The package types were not able to be resolved with the new `exports` field being added. This PR fixes that by defining the `import` and `require` properties in `exports`.

References:
- https://nodejs.org/api/packages.html#package-entry-points
- https://nodejs.org/api/packages.html#conditional-exports

Side note: I'm not sure why the `svelte` property is targeting `src/index.js` rather than `dist/index.js`. I tested using `dist` locally and my SvelteKit project compiled and ran without issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,14 @@
     "svelte": "src/index.js",
     "exports": {
         ".": {
+			"import": {
+				"types": "./dist/index.d.ts",
+				"default": "./dist/index.mjs"
+			},
+			"require": {
+				"types": "./dist/index.d.ts",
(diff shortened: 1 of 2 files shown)


isaacHagoel:
Thanks for this.
Using "dist" in the "svelte" entry caused [this issue](https://github.com/isaacHagoel/svelte-dnd-action/issues/526).
Svelte is a compiler so it wants access to the source files which does make sense (that way it can compile them together with the app, tree shake etc).
Can you please bump the version and add an entry in release-notes.md?

PR #538, opened 2024-01-26
docs: update README to help sveltekit + TS users

gurpreetatwal:
Hi there! I'm using this library with SvelteKit and ran into some issues after I followed the instructions in the README to set up types for the events. The language server started complaining that it could not find type declarations for things like `./$types` or `$lib/....`. It turns out that the Typescript does not allow "extending" of the `include` key and adding an `include` element at the root `tsconfig.json` resulted in the ts compiler not knowing about the types that sveltekit declares. Thankfully this is easily solved by using sveltekit's function to modify the generated config rather than editing the root config. 

I just threw this in the readme for now, but happy to move it to where ever you feel is more appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,7 +288,8 @@ You may need to edit `tsconfig.json` to include `global.d.ts` if it doesn't alre
 >   kit: {
 >     typescript: {
 >       config(config) {
->         config.include.push('global.d.ts');
+>          // This path is relative to the ".svelte-kit" folder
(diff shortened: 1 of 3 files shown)


isaacHagoel:
Thanks for this. Do you mind bumping the version in package.json and updating release-notes.md?

gurpreetatwal:
Done! Apologies that it took that long

isaacHagoel:
Thanks

PR #542, opened 2024-02-21
Improve custom events typing

quangthandev:
## Description:

I have added types for action `dndzone` custom events `on:consider` and `on:finalize`, which will prevent type errors and infer the type of the events properly when using TypeScript.

### Before:

Screenshot 2024-02-21 at 19 37 17

### After:

Screenshot 2024-02-21 at 20 00 16

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,9 +252,11 @@ setDebugMode(true);
 
 ### Typescript
 
-If you are using Typescript, you will need to add the following block to your `global.d.ts` (at least until [this svelte issue](https://github.com/sveltejs/language-tools/issues/431) is resolved):
+#### Setup (Optional)
 
-#### Svelte 3 or below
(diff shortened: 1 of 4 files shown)


quangthandev:
Sorry I've just re-read the TypeScript section of the docs. The reason the typings have not been added internally is support for Svelte 3 or below?

isaacHagoel:
We do need to support Svelte 3, 4 and 5.
I only played with 5 in the playground for a short amount of time and haven't tried the lib with it.
Will this solution not work with Svelte 3 and 4?

quangthandev:
I've just added support for Svelte 3 or below with event name `onconsider` and `onfinalize` in the latest commit. It works well when tested with Svelte 3, 4, and 5. I think as long as the `svelte/action` API is not changed we're good to go. Please check it out!

isaacHagoel:
Thanks. Blocking time tomorrow to review

isaacHagoel:
Looks good. Can you please bump the version in package.json and update release-notes.md?

quangthandev:
Done. Should the TypeScript section of the docs be updated too?

isaacHagoel:
yes please

quangthandev:
Done. I kept the old docs in place for reference rather than removing it.

PR #335, opened 2021-12-08
Option to disable drag in some items

Fedeorlandau:
This PR validates if the element exists on the items map and also if the item has the `dragDisabled` option.

I was wondering if we can include a playground folder with svelte to test the changes live.

How to enable this option?

```
 let board = [
        {
            id: 1,
            name: "TODO",
            items: [
                {id: 41, name: "item41"},
                {id: 42, name: "item42"},
                {id: 43, name: "item43"},
                {id: 44, name: "item44", dragDisabled: true}
            ]
        },
        {
            id: 2,
            name: "DOING",
            items: []
        },
        {
            id: 3,
            name: "DONE",
            items: []
        }
    ];```

If you add any other div that is not present on the items map it will also prevent the dragging, check this example
![drag](https://user-images.githubusercontent.com/10283686/145244233-b85cf23f-959c-408b-9155-063fa57d76d3.gif)

PR #544, opened 2024-02-25
Allow add non-draggable element in dropzone container

### LINKED ISSUE #452: Ability to include an element in a dnd zone, without it being draggable (Element-specific dragDisable flag)
I am wondering if there is a proposed way to do this, and if not, what you think of my solution.

I am trying to add elements into the dnd zone that "float" either in the beginning or end of the zone, but do not interact as dnd elements. 

I couldn't find a way without having to play around with the entire zone's disableDrag flag. 

https://github.com/isaacHagoel/svelte-dnd-action/assets/60684710/63356db9-3638-4f70-b2e1-b1e84e380f97

pointerAction

example_usage

mskocik:
Change how dndzone children are being iterated to allow non-draggable element into drozone container AFTER item array. Fix #452. 

This will be possible afterwards: 
```html
<section use:dndzone>
  {#each itemsas item(item.id)}
    <div animate:flip>
      <!-- draggable -->	     
    </div>
  {/each}
  <span data-dnd-exclude>NON DRAGGABLE</span>
</section>
```

As of this comment:

> It sounds like a can of worms of edge-cases and complexity but I am open to ideas. i am happy to look at more detailed requirements or a PR if you can provide.

_Originally posted by @isaacHagoel in https://github.com/isaacHagoel/svelte-dnd-action/issues/452#issuecomment-1734542796_

All tests passed and also all example REPLs I _tried_ continue to works.

--- 
There is similar PR (based on name) #335 for this topic, but I would say, my PR is easier to reason about 😊

### DIFF
--- a/src/keyboardAction.js
+++ b/src/keyboardAction.js
@@ -195,10 +195,10 @@ export function dndzone(node, options) {
                 e.preventDefault(); // prevent scrolling
                 e.stopPropagation();
                 const {items} = dzToConfig.get(node);
-                const children = Array.from(node.children);
+                const children = Array.from(node.children).slice(0, items.length);  // slice to prevent swap out of items range
                 const idx = children.indexOf(e.currentTarget);
                 printDebug(() => ["arrow down", idx]);
-                if (idx < children.length - 1) {
+                if (idx < items.length - 1) {
                     if (!config.autoAriaDisabled) {
                         alertToScreenReader(`Moved item ${focusedItemLabel} to position ${idx + 2} in the list ${focusedDzLabel}`);
                     }
@@ -213,7 +213,7 @@ export function dndzone(node, options) {
                 e.preventDefault(); // prevent scrolling
                 e.stopPropagation();
                 const {items} = dzToConfig.get(node);
-                const children = Array.from(node.children);
+                const children = Array.from(node.children).slice(0, items.length);  // slice to prevent swap out of items range
                 const idx = children.indexOf(e.currentTarget);
                 printDebug(() => ["arrow up", idx]);
                 if (idx > 0) {
@@ -312,7 +312,7 @@ export function dndzone(node, options) {
 
         node.addEventListener("focus", handleZoneFocus);
 
-        for (let i = 0; i < node.children.length; i++) {
+        for (let i = 0; i < config.items.length; i++) {
             const draggableEl = node.children[i];
             allDragTargets.add(draggableEl);
             draggableEl.tabIndex = isDragging ? -1 : config.zoneItemTabIndex;
--- a/src/pointerAction.js
+++ b/src/pointerAction.js
@@ -511,7 +511,7 @@ export function dndzone(node, options) {
         dzToConfig.set(node, config);
         registerDropZone(node, newType);
         const shadowElIdx = findShadowElementIdx(config.items);
-        for (let idx = 0; idx < node.children.length; idx++) {
+        for (let idx = 0; idx < config.items.length; idx++) {
             const draggableEl = node.children[idx];
             styleDraggable(draggableEl, dragDisabled);
             if (idx === shadowElIdx) {


isaacHagoel:
This seems to only allow non draggables within the container **after the last child**. 
I understood the feature request as non-draggables can be anywhere within the container and in fact had people who wanted the first item (e.g some kind of a title) or middle items to be non draggable. This is where you get into all of the edge cases I was referring to.

mskocik:
Yes, only as I wrote in the description. The related issue mentions also "before" position (which I noticed after rereading now), but I was referring to the video of desired outcome. 

When I was playing with, I did all variants (before, inside items, after). Inside items it was possible only with some custom property defined on items themselves, but it worked, even toggling it. I just didn't want to combine it.
Placing non-draggable before items was more problematic, because lib doesn't expect this and dragging and droping looked bad.

I can add ability to allow marking items as non draggable through property. But maybe that could go into another PR?

Personally I think it's better to have something than nothing

isaacHagoel:
I guess what I am struggling to understand is what is it that you can do
with this feature that you can't do without it.
Why not place the non draggable element right before the dndzone and have
both share the same parent container. With css it will look like that
element belongs in the list.
Same question applies to an element at the end or even in the middle (use 2
dndzones and place the static item in between).
I would appreciate it if you could explain.

If we do go forward with something like you suggest the implications on
aria need to be considered as well.

mskocik:
I will respond properly when I get back on Wednesday (just letting you know)

mskocik:
This [REPL](https://svelte.dev/repl/a146bf7cdd854fb4a380b79f58fcf91b?version=4.2.12) shows what's impossible to do correctly with wrapping div.

IMHO enforcing some type of markup shouldn't be required, when simple solution exists. And it exists, we just need to take into account that `node.children` don't strictly represent `items` property.

Aria shouldn't be concern at all, because "init" loop skips those non-draggable items (I am speaking in general, although this PR is only about _after_ non-draggables, which handles) so they are not participating in dnd logic in any way.
Sidenote: I am not sure aria works correctly, because aria related elements do not change their content when I focus or move some element by keyboard.

image

--- 

This PR could be extended to handle all cases (at least I can try to implement before, which was glitchy): 
- before draggables
- inside items (this one would be possible only by some special property on `item` object)
- after - already in PR

isaacHagoel:
Nice example! 
Regarding: "Aria shouldn't be concern at all", have you tried using it with a screen reader? 
In your example will the instructions explain to the user that there is a non draggable but still focusable and usable button inside the zone at the same level as the items?
Regarding the "If we can do something we should" sentiment - I am the one that will need to maintain the lib and potentially add other enhancements on top of any change that is introduced and I have to consider the long term implications of making a change such taking the items length as a source of truths instead of the number of children. Also, I want to deliver complete features and supporting only one case, in which the non draggable is at the end feels incomplete. 
Again, it's not that I am against the idea. I think that a more complete implementation would use some data attribute to mark elements as non draggable and would account for different ways it might be used (including flipping it dynamically).

mroforolhc:
> Why not place the non draggable element right before the dndzone and have both share the same parent container. With css it will look like that element belongs in the list.

I thought the same solution, but I catch bug: when I set `overflow: scroll` on the parent container and add below exactly the same container, dndzone into the first container will be overlap dndzone into the second container.

REPL: https://svelte.dev/repl/77eb045ce8044b5aac30df935492cdfa (I try to add a button under list into the scrollable container)

https://github.com/user-attachments/assets/31c2ff65-1c32-4421-8997-8f8849567d1d

isaacHagoel:
thanks @mroforolhc !
I created #609

isaacHagoel:
FYI @mroforolhc fix here: https://github.com/isaacHagoel/svelte-dnd-action/pull/610
I still need to make sure it doesn't break anything else before I merge and release but looks good so far

PR #549, opened 2024-03-16
506 - global scroller

isaacHagoel:
created a single scroller that deals with all the scrolling elements that are parents of dndzones of the relevant type, including the window. This is the basic implementation.


dusty-phillips:
I tested this in my project with nested containers and a extra header at the top and it seems to be working. 🙏🏻

isaacHagoel:
@dusty-phillips thanks for testing. I will work to finalize and merge it (aiming to merge it tomorrow).

isaacHagoel:
Should be good to go. Will do another round of testing on it tomorrow and if all good will merge

PR #551, opened 2024-03-18
540 - shallow copy of the items and...

### LINKED ISSUE #540: Items in the config are updated without waiting for the user initiated update function to be called
Items in the config are updated without waiting for the user initiated update function to be called. 
One example could be seen in this line https://github.com/isaacHagoel/svelte-dnd-action/blob/master/src/pointerAction.js#L203.
Need to check the code for other instances.
Normally it has no effect but it can result in unexpected behaviour. 
When fixing it need to check that the fix doesn't break existing working examples.

isaacHagoel:
shallow copy of the items before splicing and better handling for when there is no shallow element to animate to
fixes #540


isaacHagoel:
Fixes #553 as well

PR #556, opened 2024-04-10
fix: prevent drag from scrolling the page on safari

Draconizations:
This library currently has an issue on touch devices (at least iOS safari), where trying to drag and drop will attempt to scroll the page. I ran into this with my own project, as well as any of the example REPLs.

I was able to fix it on my own project by adding a single `e.preventDefault()` in the right spot.

Figured I might as well PR this. I don't think this should affect anything other than fixing the scrolling issue, but let me know if it does.

### DIFF
--- a/src/pointerAction.js
+++ b/src/pointerAction.js
@@ -384,6 +384,7 @@ export function dndzone(node, options) {
             printDebug(() => "cannot start a new drag before finalizing previous one");
             return;
         }
+        e.preventDefault();
         e.stopPropagation();
         const c = e.touches ? e.touches[0] : e;
         dragStartMousePosition = {x: c.clientX, y: c.clientY};


isaacHagoel:
Thanks for this. Interesting, I tested it on an iphone now and you're right.
It used to work fine (the stopPropagation right next to the line you add it used to do the trick).
The problem with preventing default is that it would prevent items from being clickable (try clicking on the items in the board [here](https://svelte.dev/repl/e2ef044af75c4b16b424b8219fb31fd9?version=3.59.2)) I think... can you check?

Draconizations:
Ah yeah. You're right about it canceling click events. It works using a mouse, but not on mobile devices.

This can be worked around with adding `e.target.click()` to `handleFalseAlarm` (if the event is a touch event). Testing this out, this doesn't cause a double click both on mobile (safari) and desktop (firefox)

src/pointerAction.js
```diff
@@ -350,11 +350,16 @@ export function dndzone(node, options) {
...
     }
-    function handleFalseAlarm() {
+    function handleFalseAlarm(e) {
         removeMaybeListeners();
         originalDragTarget = undefined;
         dragStartMousePosition = undefined;
         currentMousePosition = undefined;
+
+        // dragging initiated by touch events prevents onclick from initially firing
+        if (e.type === "touchend") {
```
isaacHagoel, line 360: cool solution but preventDefault is running regardless of whether it is a touch event or not, right? if so, this will not allow clicking with a mouse.

isaacHagoel (commented):
Please also bump the version in package.json and update release-notes.md
thanks again

src/pointerAction.js
isaacHagoel: I missed that you said you tested it with a mouse and the preventDefault doesn't stop the click if i understand you correctly. if that's the case you can ignore this comment. I can try to find time to test this properly tomorrow and if it works and the package.json and release-notes are updated i will merge

PR #567, opened 2024-04-29
Allow svelte ^5.0.0-next.0 as peer dependency

alexbehl:
When using svelte 5.0.0-next, svelte-dnd-action causes a peer dependency issue. Since svelte-dnd-action appears to be compatible with svelte 5, it would be helpful to allow svelte 5.0.0-next as peer dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "svelte-dnd-action",
     "description": "*An awesome drag and drop library for Svelte 3 and 4 (not using the browser's built-in dnd, thanks god): Rich animations, nested containers, touch support and more *",
-    "version": "0.9.43",
+    "version": "0.9.44",
     "repository": {
         "type": "git",
         "url": "git+https://github.com/isaacHagoel/svelte-dnd-action.git"
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
         "rollup-plugin-copy": "^3.3.0"
     },
     "peerDependencies": {
-        "svelte": ">=3.23.0"
+        "svelte": ">=3.23.0 || ^5.0.0-next.0"
     },
     "keywords": [
         "svelte",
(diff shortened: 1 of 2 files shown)


isaacHagoel:
Hi,
I haven't actually tested it with Svelte 5. Does it work well?
Can you pls update the version in package.json as well as the release notes?

alexbehl:
I haven't encountered any issues so far

EricSSartorius:
> I haven't encountered any issues so far

Trying to update now and I noticed that all my pages that have `overrideItemIdKeyNameBeforeInitialisingDndZones` at the top of the page now get an error saying `Uncaught Error: can only override the id key before initialising any dndzone`

It worked in 4 but doesn't work after moving to 5.

benkuper:
hello any update on that ?

isaacHagoel:
Sorry. Blocked some time to look into this tonight (Sydney time zone)

EricSSartorius:
Is this ok being merged even though `overrideItemIdKeyNameBeforeInitialisingDndZones` no longer works properly in v5?

isaacHagoel:
@EricSSartorius sorry. I missed your original comment somehow.
I didn't test this option. ideally please create an issue. 
When are you calling the override? Can you try importing and calling it at the top level before creating any components with dnd zones inside? it is possible that svelte 5 calls the action (which creates the first zone) before running your override

EricSSartorius:
> @EricSSartorius sorry. I missed your original comment somehow.
> I didn't test this option. ideally please create an issue. 
> When are you calling the override? Can you try importing and calling it at the top level before creating any components with dnd zones inside? it is possible that svelte 5 calls the action (which creates the first zone) before running your override

In v4 I am calling it at the top of my component like the example in the readme. In v5 it no longer works that way. I tried throwing it in different things like $effect.pre with no luck either. It only works after page refresh. I can make an issue 👍

PR #588, opened 2024-06-12
fix: fix issue with empty canvas

rubenfiszel:
Otherwise will throw an error, see: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage#exceptions


isaacHagoel:
Thanks! can you bump the version in package.json and add an entry to release-notes.md?

rubenfiszel:
done @isaacHagoel

isaacHagoel:
thanks

PR #600, opened 2024-09-13
Fixes wrongful element decoration

### LINKED ISSUE #596: Svelte 5: Not removing 'data-is-dnd-shadow-item-internal="true"' property so elements get hidden permanently
Hello! In Svelte 5 I am having an issue where once I release an element, it keeps the HTML property added by the library `data-is-dnd-shadow-item-internal="true"`, which means the "visibility=hidden" CSS property is kept, which means the element is still there, still occupies space but cannot be seen, loses visualization.

In Svelte 4 it works as expected.

Using the migrate script doesn't help.

In the reproducible example below, just drag the elements around and they will lose their visualization but will still exist and can be inspected via regular context click -> Inspect to see that the HTML property is still there. 

[Reproducible in Svelte 5 preview here](https://svelte-5-preview.vercel.app/#H4sIAAAAAAAAA-0aC1Mbx_mvfFESI1y9MMWND44MQYrNGAcqcFOP5WFWt99JO5x2r3t7gKzov3f2pXsgUeO2004CA0K3-71fe_d9t2jELMGsEXxcNDiZYSNoHKVpo9VQ81RfZDeYKGy0GpnIZaRXDrJIslQdjvhIsVkqpIKF_j5SlNPPgmPLXl28Oeqf_Xp1fnp0PHhzdtofDK9OLgfvrk76DuByePL69WB4UYU3MO-Ohm8Hw6vz4dn5YHj54eqXo3cDD3b2fng8uNAXS4ilmMGWlbFNOW2TSDHBt_bLwkGcsBSqwF3C2YwoNJBl2EgiUTi4Qa76LEuJiqYoa8geqYJ4K5ki4wTrjDIlpMEAAOh2oY8x4whqimC2MqBMYqSSOTAOasoyiMQsFRy5skh4Z1hEgmcKMkwwUkhPFM5OKIQrvk2eJ8n2_hoULihmx5pmhhWMj5-2rSZ8pCwodSpDuN4QTYfgwLVh-7kk2ubvMghhr-f2E1QwEzKd9lmmeWm-SuZY2o6QK4l9SSYTpGf8OJeZkBBCTJKsDEclmZSolLfjnBt_Q6aIVJpUk9EW4LaPSOykEo0KGJM8UVZ-HaobiKqRqlq4k6FqMmrxljW2U8Jpgm9x3he3vFmwZTE0m9i5xjmEYQhbA65QbsFvv0FpEba24dmzinrbG5X1rJ1jtV2Ywpm2-Ufr8oX9p39GjYgonAg5HzUCGDXGiYiuR41WBUJwpSPMAFzinQ2TGpAiEwuQ1jYYtesK79TV3vXP6mWkzpLT_Oru7-Pjo58-D0__ehfVkVIpUpSKYaaRS_LabS5mfTE7rwB9rALV9KzhkxlaqU7JGJMK8ypkRFIyZglTzj4DytQD4NTGzt9IkmNhr1HjPvxyDY0vkPfY--K_KLGuKywDAuo_K_s5kcjVCX2E8EMkD4HfF36tuNWlT616ONEviaU6leLS7Xir_O-y6x9_eZWf9nZ3CWV8Hr_7NY0n_dmOfP3qKbuesut3kF36vPy_ya7d3af8eqy8T_n1sPB_9Pwi5CnDvFueMuwLhH_KMPiqE2xv7ynDnjLsS4R_yjB4bIYRQsZPpxjAU479HnNsxD_t19qLuv9fX6OYKt0b7tU3MjQNUdOQ3pJCqAssGvGujzsl2fGUJVQir7Vxvwtqm7a1-eyZ_dJJkE_UFA6hV-_72gZsn9NjwTNGUZaasN0uHEURZpnptVNUhCXgsnMOIjbLpjls4W-I9FAhYMd-9Z3-GjUrIOEUGI8FFDnv6VrsgjDH2xPXr7VbHUNiv4AwhIpdHosNvO0gxjBXkk0mKNfw1_gFbYcSmuWOvSqx9mTcvrtcsfeNZq9DsRFD05MOQ3CTmc7bwYefzo6Gfe3AFekwBD_v6fSHR6-vLi7Pzs8H_ZW77vXk_bhA2eb3huZ7n9OfGScJ-4x_HN_XNfkybz_szHE-Hif4Pj2TF6guJeJq1rPR0ednJ79cDob_rgu1HMcJi67t8EZ7pRjg6KtOpkSqyxyZmFFTc3tfT9LO7WjHYtihm1GC8QnkKSgBqSnbgAnOkKusUOa7-hgtDIHRkiLdLpxY20ZaMqTG7eZsSSQSOl8N4jzGmrGRnchpUd9zu70KIGcUwCTDguuG0ZOmcIEW2YM4eSjkmdZ3NU58wNprPFweWLniHobQKxmiMjy0Eumw8XO0ugarQHAjw-ZWnlKijI-zre1aMBx0i1kyP3CHiJmq0nCxOlMMdJSQLAv1uZmq6cJ8LvWJ6XYCLvz5ES6-KZ0mBjnPMHDD6XCxsBncqk0vW-bEa1Wnlq0NU8qlISt45I6dcHHvJHIQsStOJQhfr5Yjboboi2-RRFN31sUsUSibxsrbEB6a5Q6j8I1OvI0j9W0gmQ2JpkPYXlpbH1B2473iht6B1lwbgjrtg_okd7n0KN7ujN-wjI0TbGt6jcp24IMyXKxNLCvQiqTgJqfCRdPlenh4vxJYHF8OHO6hJ7H4lsWWrn8SsDNVfddPGEe5tSxicaQO7CxeyxkXqyM1ZpwGxu7hwnLU35dlEB0U4aixKiGN8qYJw9BGI_wJdiqYq_j1tGllW_BSZoSLe8lZAu4e-kJcd6iRkGg18C5cVKrvj9CDANpVmYhkpJ3oBwedSpJM2tbwVbW8z9ftZWqe4D1WW5HJiwAmkoy3IKgujBmfbNWUn4k8QypueSUMSnP8tQHgsZXIo6kB_ir0a5zfY70og4xUbbRvwParIBu5VeCWJd6HB13Kbooo7rJ4Wbh2U1Cbx9uy_Q5SYKWoApuEUZ4pMWtnU0LFbVtvWpCP_-rFmk_Lw7LAlq57fi5z7ab3JDfLXqVFV9cxW9Zt7Lu6rkPGQFB24-08RTaZqgBiptqOlzPbLaNqum7Dh1QqmH6fwq2mImO2hklMiGI3lfclKLvprI5Mx3osJEUZwIv0DjKRMArjJK9gOfG_ArNTqZMeb6NKm62QEkoZnwTuWWukZkROGC-uvSxc8KoEphr1POsZ423PZafX-36_WHZS7e710jtPlETXEylyTtuRSLSx5WRMmjsvd1qws_uD--h1env6ruQ5HEGWj1WCUCCCQYTn3aqYOyuTGYovdl-2YHfX_vU6O55coiUFi1SicefCOoBeemfM_-f0zlLqtcD9FlQyS8birMiIlERMzQPodV4ZwCHSPEJw6xALCcQrhHGs79c8rpKE-zjz8L3ObgZIMmwz3ha5sswp1URmQqgpFFiOUDW-Ai5U0_lru54YJFeimhGllcf4aVPoWoxXvRa8NL_Gfo82-BrTPsJiFlTfLCZkru9C8G5dHhY3d3UrvSpi11lpteJodJ_DG3GD0ntUCZiRa30Xrm_1ELhQLELz5p7gYCqLfYnQucz7amqILGr67mywmLbWD-sstld5k6xzv2h7FqaMmH5SAK6k7N9PKUqyKVKYSJzfC43AZtOqRFXctGui9VTcoly5pxKjHXsCrt7t9D4y9UbjvmHUvsboAMdzcE2tVdYUtZmMM5HkygsirffaLwr3KZEGsLcqUCZ8YiFngc2ihCj80Gzv9b7frvq7RMIHxYuHClrJHOV7lDW2rVV452T22ZRlC9geC8_pc9vchQXwonb-uMjZZE9zuhuDXkzFbdmgt1Pk5to9xuoDhWVgyCEtvHXQ9Sdso9WYCcpihrQR6Ifv5aflPwEIG9m-4SsAAA==)

vini-brito:
This fixes #596, where the wrong element in the list is decorated as a shadow element. 

What happens is that the shadow element is not yet present when a decoration is triggered due to, for some still unknown reason, the splice at line 159 ```items.splice(shadowElIdx, 0, shadowElData);``` not running.

That splice is responsible for adding to the list of items a copy with the following properties:
 ```
"id": "id:dnd-shadow-placeholder-0000",
"isDndShadowItem": true 
```
Specifically, the ID is modified to that value and the `isDndShadowItem` property is added.
When that shadow copy is missing, we have one less item in the list and when the `decorateShadowEl(draggableEl)` function runs, it hits the correct index position but in that index position the wrong HTML element will be. 

Immediately after, this internal routine runs again, properly adds the shadow copy, then hits the correct element.
The thing is, in a healthy drag, the internal routine runs just once, in the problematic version it runs twice, first without adding the shadow copy and automatically it triggers again and properly adds the shadow copy.

Here is this quick video I first show a healthy drag, then a problematic drag:

[Screencast from 12-09-2024 21:04:17.webm](https://github.com/user-attachments/assets/50ba67f0-5f33-4c70-b270-ba0b9ad3da30)

This fix is inelegant because it doesn't get to the root cause of the issue, but apparently it works, it uses a side effect that I noticed, in the first, problematic internal routine run, one of the items object will have the ID as `"id": "id:dnd-shadow-placeholder-0000"`, and in a healthy internal routine run, the ID of the item will be the correct ID as defined by the app, so the logic here is to check if one of the items has the shadow ID, which is indicative of a problematic run, and then not to decorate the element.

Ideally, we should figure out why the problematic run happens, so any suggestion on this is appreciated.

I had a couple ideas for fixes:
1 - Undecorate the element after a problematic run.
2 - The one suggested in this PR, which is to not decorate if the side effect is noticed.

But of course I would prefer to do 3:

3 - Eventually reach the source of this problem and not even have the problematic run.

I have tons of logs I can share that allowed me to arrive at this conclusion and solution.
Example of my logs about the items, using console log stringify items, right before the decoration, showing a problematic run where this runs twice:
```
items right before decoration:
{
  "content": "Text node",
  "id": "id:dnd-shadow-placeholder-0000",
  "isDndShadowItem": true
}

{
  "content": "Text node",
  "id": "text_dOHCW0NJRU8-mqerMESzgfGCw"
}

{
  "content": "Text node",
  "id": "text_xSZJVOxc9NthNP68l2rxVP93f"
}

------ Another automatic run, this time healthy ---------

items right before decoration:
{
  "content": "Text node",
  "id": "text_AbuZBBG2dZiHu6tICxBV3YIFN",
  "isDndShadowItem": true
}

{
  "content": "Text node",
  "id": "text_dOHCW0NJRU8-mqerMESzgfGCw"
}

{
  "content": "Text node",
  "id": "text_xSZJVOxc9NthNP68l2rxVP93f"
}
```

Also, this bug was introduced in the version 42. Before that however, the last working version for my setup was 28.
From 29 to 41, all of them freeze when I try to drop with a `"getBoudingEtc..." `error.
In my setup this change doesn't seem to interfere with anything else, but I'm not sure about the wide range of use cases out there.

If it is better to discuss or further work on this before merging I'm open to it!

### DIFF
--- a/src/pointerAction.js
+++ b/src/pointerAction.js
@@ -535,7 +535,11 @@ export function dndzone(node, options) {
                     morphDraggedElementToBeLike(draggedEl, draggableEl, currentMousePosition.x, currentMousePosition.y);
                 }
                 config.transformDraggedElement(draggedEl, draggedElData, idx);
-                decorateShadowEl(draggableEl);
+
+                if (!items.some(item => item.id === "id:dnd-shadow-placeholder-0000")) {
+                    decorateShadowEl(draggableEl);
+                }
+                
                 continue;
             }
             draggableEl.removeEventListener("mousedown", elToMouseDownListener.get(draggableEl));


isaacHagoel:
Thanks for this.
Can we reproduce this issue using a minimal REPL? I'd like to try to find the root cause

vini-brito:
Sure, here is the minimal example, just one file, I removed everything I could: 

https://svelte.dev/repl/3620096a09e74af698c2633fdf5b6d00?version=4.2.18

isaacHagoel:
remind me why we can't just remove the filtering of the shadow item like I did in this REPL?
https://svelte.dev/repl/42158c219b174627980f8f6a3f0d174b?version=4.2.18

vini-brito:
It breaks nested containers. Also applying the filter is the recommendation in the docs, when I tested without, nested stuff broke.

isaacHagoel:
There are other solutions for nesting issues. In the minimal example there
is no reason for this filtering. Can we make one that justifies it or shows
the nesting issues?

vini-brito:
I made removed the filter and made several changes in userland lately and apparently this issue is gone, so the PR is unnecessary. However I cant isolate what change solved it, else I'd share it.

isaacHagoel:
Thanks for letting me know

PR #618, opened 2024-12-17
multiScroller undefined on destroy

mattcollie:
# multiScroller undefined on destroy

When rapidly moving items around within a single list, there were cases `multiScroller` was undefined when accessed (`destroy()` called) and caused an unhandled exception that left the `dndzone` in an unusable state until refresh. 

The solution I found was to wrap `multiScroller` in a truthy if block before attempting to destroy and set undefined. The check could be made safer by also ensuring `destroy` is a function on the `multiScroller` object, but that's probably overkill for this case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "svelte-dnd-action",
     "description": "*An awesome drag and drop library for Svelte 3 and 4 (not using the browser's built-in dnd, thanks god): Rich animations, nested containers, touch support and more *",
-    "version": "0.9.52",
+    "version": "0.9.53",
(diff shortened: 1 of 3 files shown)


isaacHagoel:
thanks for this.
Can you please 
1. add a comment above the change explaining why the check is there
2. bump the version in package.json
3. add an entry in relase-notes.md
?

mattcollie:
Apologies for that @isaacHagoel, I've bumped the version, added a comment and a release note.

Let me know if there's anything else you need.

isaacHagoel:
thank. merging

PR #619, opened 2024-12-20
FIX: infinite scroll on drag end

dsaisho:
Problem:
Releasing a drag element in a container WHILE it is scrolling will cause an infinite scroll in the direction that the container was scrolling.

Cause:
request animation frame keeps firing when releasing a drag item in a drop container while the drop container is actively scrolling

Solution:
Only call requestAnimationFrame in scroller.js if there is currently an active drag item.

Steps to reproduce:
-create a container for drop zone.
-create a container full of items.
-fill the drop zone container until there is an overflow.
-drag an item in the drop zone container and bring it towards the top or the bottom,  notice that the drop zone container will scroll up or down.
-release the item in the drop zone WHILE the scroll is happening.
-observe that the drop zone container scroll is stuck in the direction it was scrolling.

### DIFF
--- a/src/helpers/scroller.js
+++ b/src/helpers/scroller.js
@@ -1,4 +1,5 @@
 import {isPointInsideRect} from "./intersection";
+import { DRAGGED_ELEMENT_ID} from "../constants";
 const SCROLL_ZONE_PX = 30;
 
 /**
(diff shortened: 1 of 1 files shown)


isaacHagoel:
I am not next to a computer right now but this is weird because the scroller should be disengaged on drop (have a look at the drop handler).
Can you share the REPL that you used to reproduce this ( will save me some time)?

dsaisho:
oh man I'm a giant idiot. I was running 0.9.40. I just pulled latest and its working. I see that the 'createMultiScroller' is now returning  destroy: () => resetScrolling() which it does not in the .40 version. Im sorry for this false alarm.

isaacHagoel:
@dsaisho all good. I appreciate the fact that you tried to fix it and submitted a PR

PR #626, opened 2025-01-19
changed the logic of the shadow element retrieval when un-decorating it on drop

### LINKED ISSUE #625: Fast sorting on with optimistic updates while saving to server is broken
Moving items fast on the "Svelte dnd action - saving to  a 'server' with optimistic updates and recovery from failed operations" example sometimes breaks the whole list by removing items.

https://github.com/user-attachments/assets/28a3220a-105d-4887-b119-2cce544dcdd4

This goes aloing with a ``Uncaught Svelte error: each_key_duplicate`` error and is possibly related to handling an error response from the faked server:

image

For it being easier to reproduce I just disabled the ``alert()`` when the "server" fails. Here's my playground which is directly forked from the example without any further changes: https://svelte.dev/playground/7a5c54b988734e478b41b4e947bcc06a?version=5.17.4

I originally discovered this bug in our application but found the same happens with the basic example here.

System:
macOS 14.6.1 (23G93)
Chrome 131.0.6778.265

isaacHagoel:
fixes #625

### DIFF
--- a/src/pointerAction.js
+++ b/src/pointerAction.js
@@ -3,6 +3,7 @@ import {
     incrementActiveDropZoneCount,
     ITEM_ID_KEY,
     printDebug,
+    SHADOW_ELEMENT_ATTRIBUTE_NAME,
     SHADOW_ITEM_MARKER_PROPERTY_NAME,
     SHADOW_PLACEHOLDER_ITEM_ID,
     SOURCES,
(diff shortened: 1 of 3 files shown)


src/pointerAction.js
```diff
@@ -114,7 +115,7 @@ function unWatchDraggedElement() {
     }
     window.removeEventListener(DRAGGED_LEFT_DOCUMENT_EVENT_NAME, handleDrop);
     // ensuring multiScroller is not already destroyed before destroying
-    if(multiScroller) {
+    if (multiScroller) {
```
isaacHagoel, line 118: formatting

PR #636, opened 2025-03-09
635 - separated the adding of the dragged element to the dom from keeping the original to prevent morph from running before it's in the DOM

### LINKED ISSUE #635: Handle changes in the height of items once dragging starts
Hi, would it be possible to handle this kind of use-case https://svelte.dev/playground/0cfcc8fa9a3b4693970ad5361a997af8?version=5.22.6? The actual drag-drop doesn't work in the REPL when I use a button for some reason but it works locally. I'm using a button because I need to handle the `mousedown` event as shown in the REPL.

The reason I need to do this is to make it easier to drag through a long list of items where the items contain lot's of data that's irrelevant when items are being dragged. What's happening when the content is hidden is that the size of items changes, and the cloned dragged item is not aligned with the pointer -- especially when you drag the second or third item.

Would it be possible to account for the sizes changing when dragging starts?

isaacHagoel:
fixes #635

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "svelte-dnd-action",
     "description": "*An awesome drag and drop library for Svelte 3 and 4 (not using the browser's built-in dnd, thanks god): Rich animations, nested containers, touch support and more *",
-    "version": "0.9.57",
+    "version": "0.9.58",
     "repository": {
         "type": "git",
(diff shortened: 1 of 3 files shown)


isaacHagoel:
the code as is breaks the crazy nesting example. will need to figure out why. can't merge until i do

isaacHagoel:
i resolved the issue above and re-tested.

PR #638, opened 2025-03-10
637 - added a fix for a svelte 5 race condition that was introduced by the previous version

### LINKED ISSUE #637: version 0.9.58 breaks dndzones that store the items as a Svelte 5 $state()
This bug was introduced by #636 (that fixed #635). I tested the fix with the Svelte 3/4 syntax only and it didn't cross my mind that using $state in Svelte 5 would behave any differently but it does.
Svelte 5 users please use version 0.9.57 until this issue is fixed.

isaacHagoel:
fixes #637

### DIFF
--- a/src/pointerAction.js
+++ b/src/pointerAction.js
@@ -140,11 +140,6 @@ function handleDraggedEntered(e) {
     isDraggedOutsideOfAnyDz = false;
     // this deals with another race condition. in rare occasions (super rapid operations) the list hasn't updated yet
     items = items.filter(item => item[ITEM_ID_KEY] !== shadowElData[ITEM_ID_KEY]);
-    // this deals with svelte 5 $state that has a delay updating in some cases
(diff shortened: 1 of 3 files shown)


src/pointerAction.js
```diff
@@ -441,9 +446,9 @@ export function dndzone(node, options) {
                 originDropZoneRoot.appendChild(originalDragTarget);
                 hideElement(originalDragTarget);
                 // after the removal of the original element we can give the shadow element the original item id so that the host zone can find it and render it correctly if it does lookups by id
-                shadowElData[ITEM_ID_KEY] = draggedElData[ITEM_ID_KEY];
                 // to prevent the outline from disappearing
                 draggedEl.focus();
+                shadowElData[ITEM_ID_KEY] = draggedElData[ITEM_ID_KEY];
```
isaacHagoel, line 451: this change doesn't matter, just a leftover

PR #645, opened 2025-04-22
Fix dragHandle bug: Reset isItemsDragDisabled on click without drag

HenningHolm:
**Problem**: When clicking and releasing the drag handle without dragging, `isItemsDragDisabled` remains `false`, causing the entire element (and all others) to become draggable instead of only by the handle. I believe this is an unintended side effect that degrades the drag handle functionality.

**Solution**: Added `mouseup` and `touchend` event listeners in `dragHandle` to reset `isItemsDragDisabled` to `true` if no drag is initiated. The listeners are cleaned up properly.

**Testing**: I have tested this fix in a Svelte 5 project with both mouse and touch inputs, and it resolves the issue without introducing new problems. However, I have not conducted exhaustive testing across all use cases. Please let me know if you'd prefer this to be reported as an issue first or if further testing is needed.

### DIFF
--- a/src/wrappers/withDragHandles.js
+++ b/src/wrappers/withDragHandles.js
@@ -85,19 +85,19 @@ export function dragHandle(handle) {
         e.preventDefault();
         isItemsDragDisabled.set(false);
 
-        // Reset isItemsDragDisabled if the user releases the mouse/touch without initiating a drag
-        const onEnd = () => {
-            isItemsDragDisabled.set(true);
-            window.removeEventListener("mouseup", onEnd);
-            window.removeEventListener("touchend", onEnd);
-        };
-
-        window.addEventListener("mouseup", onEnd);
-        window.addEventListener("touchend", onEnd);
+        // Reset the startDrag/isItemsDragDisabled if the user releases the mouse/touch without initiating a drag
+        window.addEventListener("mouseup", resetStartDrag);
+        window.addEventListener("touchend", resetStartDrag);
     }
 
     function handleKeyDown(e) {
         if (e.key === "Enter" || e.key === " ") isItemsDragDisabled.set(false);
+    }
+
+    function resetStartDrag() {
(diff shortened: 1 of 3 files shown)


src/wrappers/withDragHandles.js
```diff
@@ -84,7 +84,18 @@ export function dragHandle(handle) {
         // preventing default to prevent lag on touch devices (because of the browser checking for screen scrolling)
         e.preventDefault();
         isItemsDragDisabled.set(false);
+
+        // Reset isItemsDragDisabled if the user releases the mouse/touch without initiating a drag
+        const onEnd = () => {
```
isaacHagoel, line 89: I wonder if this could simply replace the finalize function.
If not please use 'function onEnd' instead of 'const onEnd = () =>' for consistency and consider pulling it out as it manipulates the store, not a function scoped variable

isaacHagoel (commented):
Wow. I am shocked this bug went unnoticed for so log.Thanks for catching it and coming up with a fix! 👏🏻

HenningHolm:
Hi @isaacHagoel,

Thanks for the feedback! Yeah, I was wondering if I'd missed something since this bug's been around a while. 
Here's the updated PR:
- Moved the function outside "startDrag" to the same scope as "startDrag" and "handleKeyDown".
- Switched to "function resetStartDrag()" – feels clearer and matches the "startDrag" name.
- Bumped version to 0.9.61 in "package.json" and added a release note in "release-notes.md".

On "finalize": I tested removing it, and it worked fine in my setup with "dragHandleZone" since "resetStartDrag" catches all releases. But since "dragHandleZone" is used externally, I'm not sure if dropping "finalize" is safe for custom setups or complex drop zones. Feels safer to keep it, but I'm open to removing it if you're confident it’s not needed?

isaacHagoel:
@HenningHolm 
Thanks! Merging

PR #679, opened 2025-12-21
feat: add useCursorForDetection option for cursor-based zone detection

linuxsong:
## Summary

This PR adds a new option `useCursorForDetection` that uses the **cursor position** for zone collision detection instead of the **dragged element's center**.

## Problem

When dragging large elements (e.g., cards) to smaller drop targets (e.g., folder buttons), the default center-point detection can be unreliable:

- Detection only triggers when the **center** of the dragged element enters the drop zone
- If the dragged element is larger than the drop target, users may visually hover over the target but detection fails
- This is especially problematic for the **last item** in a list, where there's no adjacent zone to "catch" the drag

### Why `centreDraggedOnCursor` doesn't solve this

The existing `centreDraggedOnCursor` option **centers the dragged element on the cursor**, which:
- Changes the visual position of the dragged element (it "jumps" to the cursor)
- May feel jarring to users who expect the element to maintain its relative position

## Solution

The new `useCursorForDetection` option:
- Uses the **cursor position** for collision detection
- Keeps the dragged element at its **original visual position** (no jumping)
- Provides more **intuitive behavior** similar to macOS Finder drag-and-drop

## Example: macOS Finder Behavior

When you drag a file in macOS Finder:
- The file icon stays where you grabbed it
- But **drop detection is based on cursor position**, not the icon's center
- This allows precise targeting even when dragging large thumbnails to small folder icons

## Usage

```svelte
<div
  use:dndzone={{
    items,
    type: 'item',
    useCursorForDetection: true,  // Enable cursor-based detection
  }}
>
  ...
</div>
```

## Backward Compatibility
- Default value is false - existing behavior unchanged
- All existing tests pass
- No breaking changes

### DIFF
--- a/cypress/integration/intersection.spec.js
+++ b/cypress/integration/intersection.spec.js
@@ -1,4 +1,4 @@
-import {isCenterOfAInsideB, calcDistanceBetweenCenters, isElementOffDocument} from "../../src/helpers/intersection";
+import {isElementOffDocument} from "../../src/helpers/intersection";
 
 function makeDiv(widthPx = 50, heightPx = 50) {
     const el = document.createElement("div");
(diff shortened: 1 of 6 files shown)


isaacHagoel:
It's a good option to have. thx for the PR.
Two things:
1. Is it possible to simplify the implementation? it looks a bit convoluted. Can we just always pass a getCursorPosition into observe which could be a function or null, if null it means we use the center as before
2. In some places instead of let, if, else you could use a ternary expression
3. Please update the README, version in package.json and release-notes.md

linuxsong:
Thanks for the review and the great suggestions! I've addressed all your feedback:
1. **Simplified implementation**: `getCursorPosition` now returns document coordinates directly (with scroll offset included), so [observer.js] no longer needs to do coordinate transformation. The logic is cleaner now.
2. **Ternary expressions**: Replaced `let`, `if`, `else` patterns with ternary expressions in [listUtil.js].
3. **Documentation updated**:
   - README.md: Added `useCursorForDetection` option description
   - release-notes.md: Added v0.9.69 entry
   - package.json: Updated version to 0.9.69

All tests pass. Please take another look!

linuxsong:
Indeed, that is a much better implementation! Thanks for the improvement.

PR #685, opened 2026-06-11
fix: stop keepOriginalElementInDom rAF loop once the drag is finalized

### LINKED ISSUE #684: Zombie rAF loop in keepOriginalElementInDom() re-arms the watcher after drop: "undefined is not an object (evaluating 'e.getBoundingClientRect')"
**Version:** 0.9.69 (latest) · **Framework:** Svelte 5 (runes) · **Browsers:** seen mostly on Safari/iOS (touch), but the race itself is browser-agnostic.

### Error

```
TypeError: undefined is not an object (evaluating 'e.getBoundingClientRect')
    at getAbsoluteRect (svelte-dnd-action/dist/index.mjs:420)
    at findCenterOfElement (svelte-dnd-action/dist/index.mjs:473)
    at andNow (svelte-dnd-action/dist/index.mjs:799)
```

(Production stack via Sentry — a classroom app where students drag fast on school iPads; it fires a couple of times a week.)

### Root cause

When a pointer drag starts, `dndzone` schedules this rAF loop and **never cancels it** (`keepOriginalElementInDom`, dist/index.mjs ~1949):

```js
function keepOriginalElementInDom() {
  if (!originalDragTarget.parentElement) {
    originalDragTarget.setAttribute(ORIGINAL_DRAGGED_ITEM_MARKER_ATTRIBUTE, true);
    originDropZoneRoot.appendChild(originalDragTarget);
    // have to watch before we hide, otherwise Svelte 5 $state gets confused
    watchDraggedElement();          // starts the observe() interval
    ...
  } else {
    window.requestAnimationFrame(keepOriginalElementInDom);  // re-arms forever
  }
}
window.requestAnimationFrame(keepOriginalElementInDom);
```

The loop waits for the framework to remove the original dragged node from the DOM. But if the drag finishes **before** that ever happens (fast drag & drop where the keyed `{#each}` ends up reusing the node, so `parentElement` never becomes null during the drag), the loop just keeps polling. `handleDrop` → `cleanupPostDrop()` sets `draggedEl = undefined`, but nothing stops this rAF loop — it stays alive as a zombie.

Later — possibly seconds or minutes after the drop — the component containing that original node unmounts (in our case: the student taps "continue" and the exercise containing the zone is destroyed). The original element loses its `parentElement`, the zombie loop finally takes the first branch, and calls `watchDraggedElement()` with `draggedEl === undefined`:

`watchDraggedElement()` → `observe(undefined, …)` → `andNow()` → `getReferencePoint()` → `findCenterOfElement(undefined)` → `getAbsoluteRect(undefined)` → crash.

### Repro sketch

1. Svelte 5, keyed `{#each}` inside a `dndzone`.
2. Do a very quick drag & drop within the same zone (the drop must win the race against the framework removing the original node — with keyed each + node reuse this is common).
3. Unmount the component containing the zone (navigate away / swap content with `{#key}`).
4. The zombie rAF loop fires during unmount → crash.

Being a race, it is intermittent — but across a fleet of iPads it reproduces reliably every few days.

### Suggested fix

Bail out of the loop once the drag is finalized:

```js
function keepOriginalElementInDom() {
  if (!draggedEl) return; // drag already finalized — don't re-arm the watcher
  if (!originalDragTarget.parentElement) {
    ...
```

(Equivalently: cancel the pending rAF in `cleanupPostDrop()`.) We have been running this one-line guard in production via `patch-package` and the crash is gone, with no observable side effects on normal drag behavior.

Thanks for the library! Happy to open a PR with the guard if it looks right to you.

juanmajavaloyes:
Fixes #684 — full root-cause analysis there.

One-line guard: `keepOriginalElementInDom()` is a self re-arming rAF loop that waits for the framework to remove the original dragged node, but it is never cancelled. If the drop wins that race (common with keyed `{#each}` blocks that reuse the node, so `parentElement` never becomes null during the drag), the loop stays alive as a zombie after `cleanupPostDrop()` cleared `draggedEl`. When the node finally leaves the DOM — possibly long after the drop, e.g. when its component unmounts — the loop takes the first branch and calls `watchDraggedElement()` with `draggedEl === undefined`, crashing inside `observe()`:

```
TypeError: undefined is not an object (evaluating 'e.getBoundingClientRect')
    at getAbsoluteRect (dist/index.mjs:420)
    at findCenterOfElement (dist/index.mjs:473)
    at andNow (dist/index.mjs:799)
```

The guard bails out of the loop once the drag is finalized. We have been running exactly this change in production (school iPads, Svelte 5, heavy touch dragging) via patch-package and the crash is gone with no observable side effects on normal drag behavior.

### DIFF
--- a/src/pointerAction.js
+++ b/src/pointerAction.js
@@ -511,6 +511,11 @@ export function dndzone(node, options) {
         originDropZoneRoot.appendChild(draggedEl);
         // We will keep the original dom node in the dom because touch events keep firing on it, we want to re-add it after the framework removes it
         function keepOriginalElementInDom() {
+            // if the drag was already finalized (cleanupPostDrop cleared draggedEl), this rAF loop
+            // must not re-arm the watcher: the original element might leave the DOM long after the
+            // drop (ex: its component unmounts), which would call watchDraggedElement() with an
+            // undefined draggedEl and crash inside observe() (getBoundingClientRect of undefined)
+            if (!draggedEl) return;
             if (!originalDragTarget.parentElement) {
                 originalDragTarget.setAttribute(ORIGINAL_DRAGGED_ITEM_MARKER_ATTRIBUTE, true);
                 originDropZoneRoot.appendChild(originalDragTarget);


isaacHagoel:
thanks for this

PR #693, opened 2026-07-29
fix(keyboard): keep the dragged item's zone in sync with the consumer

mindlace:
A consumer re-render can land mid-grab and take the dragged item out of the zone it is being dragged from (an external update, a filter, a live edit from another client - keyboard grabs stay open for as long as the user takes). focusedDz kept pointing at the old zone, so tabbing to another zone spliced at index -1 and moved the last item of the old zone instead, which also made the dragged item disappear.

configure() already re-points focusedItem when it finds the dragged item in a zone it is re-rendering, but it never re-pointed focusedDz along with it. It does now, so a grab follows its item when the consumer moves it to another zone, whichever of the two zones is re-rendered first.

When no zone holds the item there is nothing left to drag. Tabbing to another zone now drops instead of splicing from a zone that no longer has it; escape, clicking away and dropping already ended the grab, and the arrow keys can't be reached with the item's element gone.

### DIFF
--- a/cypress/integration/keyboardAction.spec.js
+++ b/cypress/integration/keyboardAction.spec.js
@@ -23,4 +23,116 @@ describe("keyboardAction", () => {
         expect(dragStoppedEvents).to.equal(1);
         zone.remove();
     });
+
+    it("follows the grabbed item when a re-render moves it to another zone", () => {
+        const zoneA = document.createElement("div");
+        const itemA = document.createElement("div");
+        const itemB = document.createElement("div");
+        zoneA.appendChild(itemA);
+        zoneA.appendChild(itemB);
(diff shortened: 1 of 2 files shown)


isaacHagoel:
Thanks @mindlace — the stale `focusedDz` problem is real, and this PR identified the key failure mode.

I opened #694 as a follow-up that keeps the core ownership-sync idea and also handles two related invariants: item ids may repeat across different zone types, and changing the focused zone must refresh the active keyboard tab stops. It also adds isolated regression coverage for both consumer update orders, item removal, DOM replacement/reparenting, and cross-type duplicate ids, plus the 0.9.75 version bump and release notes.

The full suite is passing (39/39).

PR #698, opened 2026-07-31
feat(a11y): setAriaStrings for localizable screen-reader copy

mindlace:
### The problem

Today the only way to change the screen-reader wording is `autoAriaDisabled`, which is all-or-nothing — it turns off the aria attributes, roles and instructions along with the alerts, so a consumer who just wants French has to re-implement the whole accessibility layer via `consider`/`finalize`.

### The change

A new exported `setAriaStrings(overrides)` that merges over the string table the library speaks from:

```javascript
import {setAriaStrings} from "svelte-dnd-action";

setAriaStrings({
    dropped: ({itemLabel, zoneLabel, position, count}) => `${itemLabel} déposé dans ${zoneLabel}, ${position} sur ${count}`
});
```

- Seven keys, all optional — anything you leave out keeps its English default. The five message keys (`dragStarted`, `movedToPosition`, `movedToZoneEnd`, `movedToZoneStart`, `dropped`) are **functions** taking a context object, so consumers control word order and pluralisation rather than filling in blanks in an English sentence. The two instruction keys (`zoneActiveInstruction`, `zoneDragDisabledInstruction`) are plain strings.
- **Every message key receives the same context**: `itemLabel` and `zoneLabel` (from the `aria-label`s the consumer already provides), plus `position` and `count` — the item's 1-based seat in the zone the message is about, and how many items that zone holds. `dragStarted` additionally gets `canMoveBetweenZones`. The built-in English strings don't interpolate every field, but they are all supplied so any message can be worded positionally. Without that, a consumer can't say `Picked up Card A. To do, 1 of 5` on a grab, or name where an item came to rest on a drop.
- Callable at any time, including on a locale change - the static instruction divs already in the DOM are re-rendered. `setAriaStrings(null)` restores the built-in English.
- Validates the whole overrides object before applying any of it, so an unknown key or a wrong-typed value throws immediately and never leaves a half-applied table behind. Follows the error-message style of `setDebugMode`/`overrideItemIdKeyNameBeforeInitialisingDndZones`.
- A consumer formatter that throws is contained: `announceToScreenReader` falls back to the built-in default for that key and reports via `printDebug`. These run inside the drag lifecycle (e.g. `handleDrop`), so letting an exception escape would strand the library mid-drag.
- The instruction divs now render via `textContent` instead of `innerHTML`, since the text can now come from the consumer.

Behaviour is unchanged for anyone who doesn't call it — the defaults are the current strings verbatim, and the two callsites that gained context (`dragStarted`, `dropped`) read the seat off the same live items array the move messages already used.

### Docs & types

README gets a "Translating the screen-reader messages" section under accessibility, and the `autoAriaDisabled` row points at `setAriaStrings` for the wording-only case. `AriaStrings` and `setAriaStrings` are added to `typings/index.d.ts`.

### Tests

New `cypress/integration/aria.spec.js` (11 tests) covering defaults, partial merge, re-render of live instructions, reset via `null`, each validation error, and the throwing-formatter fallback. `keyboardAction.spec.js` gains coverage that the keyboard paths route through the table — including both branches of the cross-zone announcement (end vs. beginning), the `canMoveBetweenZones` branch of `dragStarted`, and that the grab and the drop are handed the item's real seat and the zone's real size. That last one grabs the middle card of a three-card zone on purpose: in a single-item zone a hardcoded position of `1` would pass either way.

`yarn lint`, `yarn build` and the full `yarn test` suite (60 passing) are green on top of current master.

### DIFF
--- a/src/helpers/aria.js
+++ b/src/helpers/aria.js
@@ -115,9 +115,11 @@ export function alertToScreenReader(txt) {
 }
 
 /**
- * Overrides the strings the library speaks to screen readers. Merges over the current strings, so you
- * can translate one message without restating the rest. Can be called at any time - including again on
- * a locale change, which also re-renders the static instructions that are already in the DOM.
+ * Overrides the strings the library speaks to screen readers. Merges over the built-in English defaults,
+ * so you can translate one message without restating the rest. Each call describes a whole locale rather
+ * than patching the previous one - anything a call leaves out goes back to English, so switching between
+ * two partial locales can't leave keys behind speaking the old language. Can be called at any time -
+ * including again on a locale change, which also re-renders the static instructions already in the DOM.
  * This is global and applies to all dndzones.
  * Pass null to restore the built-in English strings.
  * @param {Object | null} overrides - any subset of: dragStarted, movedToPosition, movedToZoneEnd,
@@ -145,7 +147,7 @@ export function setAriaStrings(overrides) {
                 throw new Error(`${key} should be a string but instead it is a ${typeof value}, ${toString(value)}`);
             }
         });
-        ariaStrings = {...ariaStrings, ...overrides};
+        ariaStrings = {...DEFAULT_ARIA_STRINGS, ...overrides};
     }
     if (isOnServer) return;
     Object.entries(INSTRUCTION_ID_TO_STRING_KEY).forEach(([id, key]) => {
--- a/src/helpers/aria.js
+++ b/src/helpers/aria.js
@@ -4,10 +4,24 @@ const INSTRUCTION_IDs = {
     DND_ZONE_ACTIVE: "dnd-zone-active",
     DND_ZONE_DRAG_DISABLED: "dnd-zone-drag-disabled"
 };
-const ID_TO_INSTRUCTION = {
-    [INSTRUCTION_IDs.DND_ZONE_ACTIVE]: "Tab to one the items and press space-bar or enter to start dragging it",
-    [INSTRUCTION_IDs.DND_ZONE_DRAG_DISABLED]: "This is a disabled drag and drop list"
+const INSTRUCTION_ID_TO_STRING_KEY = {
+    [INSTRUCTION_IDs.DND_ZONE_ACTIVE]: "zoneActiveInstruction",
+    [INSTRUCTION_IDs.DND_ZONE_DRAG_DISABLED]: "zoneDragDisabledInstruction"
 };
+
+const DEFAULT_ARIA_STRINGS = {
+    dragStarted: ({itemLabel, zoneLabel, canMoveBetweenZones}) =>
+        `Started dragging item ${itemLabel}. Use the arrow keys to move it within its list ${zoneLabel}` +
+        (canMoveBetweenZones ? ", or tab to another list in order to move the item into it" : ""),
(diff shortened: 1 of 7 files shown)


src/helpers/aria.js
```diff
@@ -88,3 +113,73 @@ export function alertToScreenReader(txt) {
...
+            }
+            const value = overrides[key];
+            if (FUNCTION_ARIA_STRING_KEYS.includes(key) && typeof value !== "function") {
+                throw new Error(`${key} should be a function but instead it is a ${typeof value}, ${toString(value)}`);
+            }
+            if (STRING_ARIA_STRING_KEYS.includes(key) && typeof value !== "string") {
+                throw new Error(`${key} should be a string but instead it is a ${typeof value}, ${toString(value)}`);
+            }
+        });
+        ariaStrings = {...ariaStrings, ...overrides};
```
isaacHagoel: The README and PR description say omitted keys keep their English defaults, and that `setAriaStrings` can be called again when the locale changes. Because this merges over the current `ariaStrings`, switching between two partial locale objects leaks translations from the previous locale. I reproduced this with a French `dropped` override followed by a German instruction-only override: drops remained French instead of returning to the documented English default. Could we merge each non-null call over the defaults and add a regression test that switches between partial locales with disjoint keys?

```suggestion
        ariaStrings = {...DEFAULT_ARIA_STRINGS, ...overrides};
```

isaacHagoel (commented):
One locale-switching issue surfaced in focused testing; details and a suggested fix are inline.

src/helpers/aria.js
mindlace: Confirmed and applied your suggestion in 0b50b18 — each call now merges over `DEFAULT_ARIA_STRINGS`.

The regression coverage went into the existing `can be called again to switch locale` test rather than alongside it; it's now `treats each call as a whole locale, not a patch on the previous one`. A locale switch involves three kinds of key, and one test covers all of them: `dropped` is named by both locales (the new one wins), `movedToZoneEnd` only by the new one (it lands), and `zoneActiveInstruction` only by the old one (back to English). Only that last kind exposes the leak, and it doubles as coverage that the live instruction div re-renders on the reset. Verified it fails on the old line and passes on the new one.

The old version of that test overrode the same key in both calls, so it passed either way — that's the gap you found. Once the disjoint case was covered, same-key replacement couldn't fail independently of it, so keeping both would have been two tests for one contract. Test count is unchanged at 60.

Two things beyond the one-liner, since the leak was also a documentation bug: the JSDoc on `setAriaStrings` and the doc comment in `typings/index.d.ts` both said "merges over the current strings", which described the behaviour you're removing. Both now say the overrides merge over the English defaults and that each call describes a whole locale rather than patching the previous one. The README already promised "what you leave out keeps its English default" (true only after this fix), so I left that line alone and added the whole-locale semantics to the paragraph about calling it again on a language change.

`yarn lint`, `yarn build` and the full suite are green.

(Force-pushed once to squash the fix and its test into the single commit above — the tree is identical to what was there a moment ago, only the history is tidier.)

isaacHagoel (approved):
Reviewed the updated head 0b50b18c. The locale-switching fix and regression coverage address the prior finding. Full Cypress suite passes (60/60), and yarn build passes. No remaining findings.

Issue #511, opened 2023-11-04
possible to disable keyboard dragging (or better assign different keys)?


ulaas:
hello, thanks for the great work.
Is there a way to disable keyboard drag and only use pointer as a SOURCE or at least change the Space or Enter keys so i can use them for other purposes?

isaacHagoel:
disabling keyboard drag would automatically make you app inaccessible
currently the lib doesn't provide a way to reconfigure the keys
what is your usecase? what are you trying to do?

ulaas:
i am checking the lib to see if i can use it for a playlist implementation with fully customized keyboard navgation. basically i would want enter and/or space key to trigger the playback of the focused item. nevertheless, i understand your stance about acessibility. however ability to reassign different shortcuts (say with a modifier) for dragstart whilst keeping the lib defaults as they are may not be so nasty right?. either way if i decide to use the lib it wont stop me from using it via patching just a little extra maintenance on my side. let me know what do you think about overriding shortcuts. again, real nice work! thanks.

isaacHagoel:
sorry for the delay. I am open to a PR that would add customisations of that sort if you're interested in implementing

ulaas:
no problem, i will let you know if i happen to provide an elegant solution. not focused at it now but surely i will contribute back if you are ok with the idea. you can close this issue i guess. Thank you very much!

mskocik:
> what is your usecase?

Keyboard dragging is somewhat broken by design. It allows to "exit" drag without any event being dispatched. Like [this REPL](https://svelte.dev/repl/3420ec01899a4161be49bddabbc4dd08?version=4.2.12). And if you rely on `consider`/`finalize` events it can break things. Would be easier to prevent dnd by keyboard in this case.

I believe that accessible keyboard drag-n-drop should work like this: 
- only drozones should be focusable (to allow fast navigation through DOM)
- "enter" it by pressing <kbd>space</kbd> or <kbd>enter</kbd> when focused.
- after that you can navigate by arrows through the items without dragging them, ie. 'selection' phase.
- when you find item you want to move, you press <kbd>Enter</kbd>, triggering drag start,
- by hitting <kbd>Enter</kbd> again, item is "dropped" at current position (`finalize` dispatched). 
- moving items from one dropzone to another should be possible just by arrow keys when you reach end on one dz, no by <kbd>Tab</kbd>, which is totally unexpected.

[...]

PR #701, opened 2026-08-02
feat: add onActivate to hand Enter to the consumer (#511)

mindlace:
Adds an optional `onActivate` so a consumer can take the _Enter_ key for their own use without giving up keyboard drag and drop.

When `onActivate` is supplied, pressing _Enter_ on an item that is not being dragged calls it with that item's id instead of starting a drag. _Space_ still starts a drag, so the zone stays keyboard-accessible. Omit the option and _Enter_ behaves exactly as it does today.

```js
use:dndzone={{items, onActivate: id => openItem(id)}}
```

### Why

#511 asks for a way to reassign or free up _Space_/_Enter_. The concrete case there is a playlist where _Enter_ should play the focused item; you replied that you were open to a PR adding customisation of that sort.

The part that seems worth fixing is what people do in the meantime. In that thread the reported workarounds are setting `tabindex="-1"` on every zone and item and wrapping the whole thing in custom listeners, or dropping `keyboardAction.js` altogether. Both are strictly less accessible than the library's own handling. This option removes the reason to reach for them: the consumer gets _Enter_, and the library keeps announcing, managing focus and driving the drag from _Space_.

It is deliberately not a general key-remapping API. _Enter_ is the one key with a genuine conflict, because it is the conventional "activate this thing" key and is doubled with _Space_ here — so handing it over costs no capability. I am happy to take this further toward configurable keys if you would prefer that shape.

### Implementation notes

- The check sits inside the existing shared `Enter`/`" "` case rather than splitting `Enter` into its own case, so the nested-interactive-element guard, `preventDefault` and `stopPropagation` all keep applying unchanged. The diff is one guard clause.
- It reads the item id from the zone's `items` rather than going through `setCurrentFocusedItem`, so an activation leaves the module's drag state untouched — no `focusedItem`/`focusedItemId` set while `isDragging` is false.
- `onActivate` is threaded through `configure()` like the other options, so it can be added or removed by a later update.
- `src/action.js` gains the option in `validateOptions`' destructure so it is not rejected as unknown.

### Tests

`cypress/integration/keyboardOnActivate.spec.js`, 9 cases covering both directions:

- with the option: _Enter_ activates and does not grab; _Space_ still grabs; _Enter_ still drops an in-progress drag
- without it: _Enter_ grabs, and drops an in-progress drag, as before
- the handler being added and removed through `update()`
- a nested `<button>` keeping its own _Enter_
- `dragDisabled` suppressing activation

`yarn lint` and the full `yarn test` suite pass (69 tests).

Docs: new row in the options table plus a note on the existing keyboard-support bullet, and the type in `typings/index.d.ts`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ An options-object with the following attributes:
 | `centreDraggedOnCursor` | Boolean | No | `false` | Setting it to true will cause elements from this dnd-zone to position their center on the cursor on drag start, effectively turning the cursor to the focal point that triggers all the dnd events (ex: entering another zone). Useful for dnd-zones with large items that can be dragged over small items. |
 | `useCursorForDetection` | Boolean | No | `false` | Setting it to true will use the cursor position instead of the dragged element's center for drop zone detection. This improves accuracy when dragging large elements over small drop targets. Unlike `centreDraggedOnCursor`, this option does not reposition the dragged element. |
 | `dropAnimationDisabled` | Boolean | No | `false` | Setting it to true will disable the animation of the dropped element to its final place. |
+| `onActivate` | Function | No | `undefined` | Called with the item's id when _Enter_ is pressed on an item that is not being dragged.<br />Signature:<br />function(itemId) {}<br /><br />Supplying it hands _Enter_ to your app - for example to open the item - instead of starting a keyboard drag. The _Space_ key still starts a drag, so the zone remains keyboard-accessible. Omit it and _Enter_ keeps its default grab/drop behaviour. |
 | `delayTouchStart` | Boolean \| Number | No | `false` | Prevents accidental drags on touch devices that should have been scrolls.  
  • `true` – enable with sensible default (80 ms).  
  • `number` – custom delay in milliseconds.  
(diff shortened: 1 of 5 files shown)


isaacHagoel:
Thanks. Before settling on the API, could you describe the concrete application behavior you need and why an interactive child such as a button or link doesn’t cover it? I’m concerned that onActivate makes the drag action responsible for application-level activation.

mindlace:
In my case, it's a kanban-type app, and enter is for "make this selected card become editable", while space is "make this card draggable". I could have a dedicated edit button inside the card, but clicking on the text makes it become editable (not a button) so adding a button for keyboard edit felt excessive.

mindlace:
Would an approach that deregistered Enter with svelte-dnd-action be preferable?

isaacHagoel:
Thanks, that makes sense. Would a more general option like this solve your use case?

```ts
keyboardDragTrigger?: "space" | "enter" | "space_or_enter"
```

It would default to `"space_or_enter"` to preserve the current behaviour. Setting it to `"space"` would mean that only Space is handled by the drag action, while Enter is left completely untouched for your application to use to make the card editable.

Would that give you what you need, or is there a reason you need the drag action itself to invoke `onActivate`?

mindlace:
Yes - keyboardDragTrigger covers my use case, and I agree it is the better shape: the library stops claiming the key instead of brokering the activation. I've opened #702 implementing the API you described (default "space_or_enter"; "space" leaves Enter completely untouched). Closing this one in its favour.

